Issue #25, opened 2018-08-22
Create template for documents that describe Custom Resources


derberg:
Confirm these statements before you submit the issue:

- [x] I have searched open and closed issues for duplicates.
- [x] I have read the contributing guidelines.
---

**Description**

- Prepare a template and store in one location with other documentation templates in guidelines directory
- Base the template on already existing documents:
  - https://github.com/kyma-project/kyma/blob/master/docs/api-gateway/docs/011-api-custom-resource.md
  - https://github.com/kyma-project/kyma/blob/master/docs/kyma/docs/040-installation-custom-resource.md
- Be consistent with already existing templates: https://github.com/kyma-project/community/blob/master/guidelines/templates/resources/cli-reference.md or suggest refactor that will have to be done also in other templates

**Reasons**

Every documentation topic must follow https://github.com/kyma-project/community/blob/master/guidelines/content-guidelines/content-strategy.md#documentation-types and all those types must have templates which make it much easier for anybody to contribute content

PR #35, opened 2018-08-29
#25: Add a template for documents that describe Custom Resources

klaudiagrz:


guidelines/templates/resources/custom-resource.md
```diff
@@ -0,0 +1,36 @@
...
+
+The {CRD full name with the domain} Custom Resource Definition (CRD) is a detailed description of the kind of data and the format used to {rest of the CRD description}. To get the up-to-date CRD and show the output in the yaml format, run this command:
+
+```
+kubectl get crd {CRD full name with the domain} -o yaml
+```
+
+## Sample Custom Resource
+
+This is a sample CR that {description of the CR functions}.
```
tomekpapiernik: Instead of this sentence, I think there should be a general explanation (block-quote format) telling the user what needs to go into this section and how to describe it. 

The section should contain a Custom Resource created using the CRD mentioned in the introductory section. 
The CR should be a valid one. If this is impossible, it should be as close to being valid as it is possible. 
Its functionality should be described, stressing any options used. See the examples below, the "options" are in bold.

> This is a sample CR that the API-Controller listens for to expose a service. **This example has the authentication section specified which makes the API Controller create an Istio Authentication Policy for this service.** 

You don't have to specify this section. If you don't, the service doesn't get an Istio Auth Policy. 

> This is a sample CR that controls the Kyma installer. **This example has the action label set to install**, which means that it triggers the installation of Kyma.

You could set the label to **uninstall** and get a different result when you apply the CR to the cluster.

guidelines/templates/resources/custom-resource.md
```diff
@@ -0,0 +1,36 @@
+---
+title: CRD name
```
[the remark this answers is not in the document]
klaudiagrz: I've tried to put CRD name into { } but this is how it's displayed in the metadata filed:
2018-08-30_15-24-25 copy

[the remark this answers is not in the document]
derberg: this is the name always, taken from spec.names.singular
https://github.com/kyma-project/kyma/blob/master/resources/cluster-essentials/templates/service-binding-usage.crd.yaml#L11

I think we do not have to worry about how it is displayed, consistency with pointing out the fields that user of the template has to change is more important in my opinion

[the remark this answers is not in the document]
derberg: so that would be `CRD singular spec name`?

guidelines/templates/resources/custom-resource.md
```diff
@@ -0,0 +1,36 @@
...
+type: Custom Resource
+---
+
+> This document is a ready-to-use template for a Custom Resource (CR) document that provides the sample Custom Resource and description of its elements for a given CRD. For details, see the existing CR documents for the [Installer](https://github.com/kyma-project/kyma/blob/master/docs/kyma/docs/040-installation-custom-resource.md) and the [Api-gateway](https://github.com/kyma-project/kyma/blob/master/docs/api-gateway/docs/011-api-custom-resource.md).
+
+> **NOTE:** Blockquotes in this document provide instructions. Remove them from the final document.
+
+
+The {CRD full name with the domain} Custom Resource Definition (CRD) is a detailed description of the kind of data and the format used to {rest of the CRD description}. To get the up-to-date CRD and show the output in the yaml format, run this command:
```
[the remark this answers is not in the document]
derberg: I told Klaudia to do it this way, but yeah you are right, this is the CDR name

guidelines/templates/resources/custom-resource.md
```diff
@@ -0,0 +1,36 @@
...
+
+```
+kubectl get crd {CRD full name with the domain} -o yaml
```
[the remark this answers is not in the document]
derberg: I told Klaudia to do it this way, but yeah you are right, this is the CDR name

guidelines/templates/resources/custom-resource.md
derberg: what you are suggesting is that templates:
- addresses that we want content provider to explain what example shows
- addresses option that content provider can add more then one example
- and that above is explained in a quote

I agree with above

guidelines/templates/resources/custom-resource.md
```diff
@@ -0,0 +1,36 @@
...
+apiVersion:
+kind:
+metadata:
+  name:
+{another_field}:
+```
+
+This table analyses the elements of the sample CR and the information it contains:
+
+> In the table, describe CR's fields starting with the **metadata.name**.
```
derberg: you are to specific here, there are more different `metadata` fields possible, same with `spec`
`In the table, describe CRD's fields starting with the **metadata** or **spec**.` 

And maybe specify that `metadata.name` is important only if some functionality of some controller depends on it. For example that the name must be the same as "blabla". (this is for example the case with CRD, the name of CRD is a combination of some specific fields and cannot be different). But if the `metadata.name` has no special requirements you do not have to really document it

guidelines/templates/resources/custom-resource.md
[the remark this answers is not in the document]
derberg: actually kind (as it is, without spaces), not singular, sorry
https://github.com/kyma-project/kyma/blob/master/resources/cluster-essentials/templates/service-binding-usage.crd.yaml#L12
as later this is the word user uses in the CR definition as a `kind: ServiceBindingUsage `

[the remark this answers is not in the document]
klaudiagrz: Fixed.

guidelines/templates/resources/custom-resource.md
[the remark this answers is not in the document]
klaudiagrz: Changed to {CRD name}.

guidelines/templates/resources/custom-resource.md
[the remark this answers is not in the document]
klaudiagrz: Changed to {CRD name}.

guidelines/templates/resources/custom-resource.md
klaudiagrz: Discussed yesterday.

guidelines/templates/resources/custom-resource.md
```diff
@@ -0,0 +1,37 @@
...
+
+The {CRD name} Custom Resource Definition (CRD) is a detailed description of the kind of data and the format used to {rest of the CRD description}. To get the up-to-date CRD and show the output in the yaml format, run this command:
+
+```
+kubectl get crd {CRD name} -o yaml
+```
+
+## Sample Custom Resource
+
+> In this section, describe a Custom Resource created using the CRD mentioned in the introductory section. Describe its functionality, stressing all the options used. Provide the Custom Resource code sample in the applicable, ready-to-use format.
```
tomekpapiernik: In this section, provide an example Custom Resource created based on the CRD described in the introductory section. Describe the functionality of the CR and highlight all of the optional elements and the way they are utilized. 
Provide the Custom Resource code sample in the ready-to-use format.


That's how I'd do it.

guidelines/templates/resources/custom-resource.md
```diff
@@ -0,0 +1,37 @@
+---
+title: {CRD kind}
+type: Custom Resource
+---
+
+> This document is a ready-to-use template for a Custom Resource (CR) document that provides the sample Custom Resource and description of its elements for a given CRD. For details, see the existing CR documents for the [Installer](https://github.com/kyma-project/kyma/blob/master/docs/kyma/docs/040-installation-custom-resource.md) and the [Api-gateway](https://github.com/kyma-project/kyma/blob/master/docs/api-gateway/docs/011-api-custom-resource.md).
```
tomekpapiernik: This document is a ready-to-use template for a Custom Resource (CR) document that provides a sample Custom Resource and description of its elements. Additionally, the document points to the Custom Resource Definition (CRD) used to create CRs of the given kind. For reference, see the existing documents for the [Installation] CR and the [Api] CR.

klaudiagrz: Applied with the plural (CRs) at the end, as we point to two custom resources.

guidelines/templates/resources/custom-resource.md
klaudiagrz: Applied.

guidelines/templates/resources/custom-resource.md
```diff
@@ -0,0 +1,40 @@
+---
+title: {CRD kind}
+type: Custom Resource
+---
+
+> This document is a ready-to-use template for a custom resource (CR) document that provides a sample custom resource and description of its elements. Additionally, the document points to the Custom Resource Definition (CRD) used to create CRs of the given kind. For reference, see the existing documents for the [Installation](https://github.com/kyma-project/kyma/blob/master/docs/kyma/docs/040-installation-custom-resource.md) and the [Api](https://github.com/kyma-project/kyma/blob/master/docs/api-gateway/docs/011-api-custom-resource.md) CRs.
```
tomekpapiernik: custom resource -> capitalize 

:D

guidelines/templates/resources/custom-resource.md
```diff
@@ -0,0 +1,40 @@
...
+The {CRD name} Custom Resource Definition (CRD) is a detailed description of the kind of data and the format used to {rest of the CRD description}. To get the up-to-date CRD and show the output in the yaml format, run this command:
+
+```
+kubectl get crd {CRD name} -o yaml
+```
+
+## Sample custom resource
+
+> In this section, provide an example custom resource created based on the CRD described in the introductory section. Describe the functionality of the CR and highlight all of the optional elements and the way they are utilized.
+Provide the custom resource code sample in the ready-to-use format.
```
tomekpapiernik: I might be picky, but why "THE ready-to-use format"? Is there a SINGLE format that is "ready-to-use"? I think this should have an indefinite article.

guidelines/templates/README.md
```diff
@@ -15,6 +15,7 @@ The table provides an overview of the template names, the repository they refer
 | [`community`](https://github.com/kyma-project/community) | [`ADR.md`](./resources/ADR.md) | Use the template for ADRs.
 | All | [`README.md`](../../guidelines/internal-guidelines/repository-template/template/README.md) | Use the template for the main README.md document in any repository within the Kyma organization. |
 | [`kyma/docs`](https://github.com/kyma-project/kyma/tree/master/docs) | [`cli-reference.md`](./resources/cli-reference.md) | Use the template to describe the syntax and the use of CLI commands for a given component. |
+| [`kyma/docs`](https://github.com/kyma-project/kyma/tree/master/docs) | [`custom-resource.md`](./resources/custom-resource.md) | Use the template to describe a custom resource (CR) and its elements. |
```
tomekpapiernik: Capitalize "custom resource"

derberg: It was my advice to not capitalize as it is not an invented name (however `nazwa wlasna` is called in english in this context). Is is just a resource, but custom one. Custom Resource Definition is an invented name
https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/

guidelines/templates/resources/custom-resource.md
derberg: It was my advice to not capitalize as it is not an invented name (however `nazwa wlasna` is called in english in this context). Is is just a resource, but custom one. Custom Resource Definition is an invented name
https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/

guidelines/templates/README.md
derberg: Sure, makes sense, as long as you start capitalizing `Resource` ;)

klaudiagrz: Ok, I'd stay with "custom resource" without capitalization. As all changes are applied and/or addressed, I'm finally merging this PR 😉

PR #59, opened 2018-09-07
Update issues-workflow.md with SIG/WG custom labels 

natasab:
**Description**

Add custom label referencing SIGs or WGs in issues.


issues-workflow.md
```diff
@@ -27,6 +27,7 @@ Here are the custom labels introduced by the Kyma team:
 * `priority/critical` indicates the top-priority of a given issue.
 * `area/{CAPABILITY_NAME}` indicates which capabilities are related to a given issue. You can assign more than one `area` label an issue.
 * `security/{SCORE}` indicates the security issue based on the [CVSSv3](https://www.first.org/cvss/calculator/3.0) security rating.
+* `{SIG-WG-LABEL}` indicates which SIG or WG identified the issue and is responsible for further follow-up of the issue.
```
mjasinski5: shouldnt be `sig/{sig-name}`? e.g `sig/core`?

natasab: @mjasinski5 was thinking about it. We have 2 concepts SIGs and WGs. 
So we get to 

*sig/core
wg/xyz*

or we simply have a standard for SIG/WG short names (as we implicitly have) and we only use the name such as:

sig-core
wg-xyz

Personally, both are fine, the second is closer to naming as we use it in slack channels, DLs etc.

derberg: naming first of all must be consistent with the labels naming we already have, so it should be:
- sig/core
- wg/consolidation

mjasinski5: So my suggestion is to have 2 separate points:
* sig/{sig-name} - e.g `sig/core`
* wg/{wg-name} - e.g `wg/consolidation`

as `{SIG-WG-LABEL}` doesnt tell the user how it should be named

natasab:
@derberg related to the comment about SIG/WG and the triage area one approach would be:

* new issue already comes with the SIG/WG label (as it is issued by a SIG/WG)
* after the area dedication, we ask "what SIG/WG?" - it would come under triage stage "SIG/WG", "Assign existing SIG/WG responsible for this issue" 

Does it make sense to you?

derberg:
@natasab sorry I didn't get the question, I think
for me it is new stage, between Help and Security

natasab:
@derberg OK, yes, i did not probably put the question right. 
-  it is a new stage, named "SIG/WG" 
- i though originally should be after Area, because Area is closely related to a SIG/WG too :)  - although i like it between Help/Security too 

Otherwise, by default, new issues coming from SIG/WG should already have the label set, but this is not modeled.

natasab:
@derberg this PR is no longer WIP.

natasab:
As agreed with @derberg kyma-triage.html is deleted completely as it is not needed. The .svg image does not need additional .html representation or file. 
@derberg please check and approve as appropriate.

natasab:
@derberg done.. i think headers had value such as "Working group (WG)"  - and the link validator seems as does not allow "()" in the header title. So let'S see if this was the problem :)

issues-workflow.md
```diff
@@ -27,6 +27,8 @@ Here are the custom labels introduced by the Kyma team:
 * `priority/critical` indicates the top-priority of a given issue.
 * `area/{CAPABILITY_NAME}` indicates which capabilities are related to a given issue. You can assign more than one `area` label an issue.
 * `security/{SCORE}` indicates the security issue based on the [CVSSv3](https://www.first.org/cvss/calculator/3.0) security rating.
+* `sig/{SIG_NAME}` indicates [which Special interest group](./sig-and-wg/README.md#special-interest-groups-and-working-groups) identified the issue and is responsible for further follow-up of the issue.
```
kazydek: ``* `sig/{SIG_NAME}` indicates which [Special Interest Group](./sig-and-wg/README.md#special-interest-groups-and-working-groups) identified the issue and is responsible for the further follow-up on the issue.``

issues-workflow.md
```diff
@@ -27,6 +27,8 @@ Here are the custom labels introduced by the Kyma team:
...
+* `wg/{WG_NAME}` indicates [which Working group](./sig-and-wg/README.md#special-interest-groups-and-working-groups) identified the issue and is responsible for further follow-up of the issue.
```
kazydek: `` * `wg/{WG_NAME}` indicates which [Working Group](./sig-and-wg/README.md#special-interest-groups-and-working-groups) identified the issue and is responsible for the further follow-up on the issue.``

derberg:
will merge as administrator as @natasab is at a conference and this PR must be merged
it is missing cla agreement but we know Natasha has it and is is just a temporary issue with cla assistant

PR #67, opened 2018-09-12
updated kyma security email address

agolawska:
updated kyma security email address to kyma-security@googlegroups.com


agolawska:
security email updated

PR #92, opened 2018-10-04
Add a proposal for feature introduction through feature gates and installer profiles

Abd4llA:
**Description**

Add a proposal for feature introduction in Kyma and use Knative as a use case.

- Add feature gates proposal.
- Add installation profiles
- Provide a rebase on Knative as a use case.


sig-and-wg/sig-core/decisions/dr-020-Feature_gates.md
```diff
@@ -0,0 +1,43 @@
...
+
+Kyma release strategy and branching model defines master should always be in a
+"releasable" state and Kyma follows a cactus branching model with releases
+branches. So in principal Kyma is following a trunk based development model with
+no long living feature branches. A core technique of applying trunk based
+development is using feature toggles, and in Kyma case introducing Kyma wide
+feature gates is needed and not just per component.
+
+This way when introducing major changes such as rebasing Kyma on Knative we can
+hide all modifications behind a kyma wide feature flag.
```
Tomasz-Smelcerz-SAP: For now this proposal only shows one use-case: installation on knative.
Proposed solution assumes a non-trivial implementation effort: building an API, golang library, etc.
If this is only to support one use case then, for me, it's not justifiable effort and we're good with current solution: helm global value to control conditional behaviour, switching components on the list.
I would even say, that we can go with current approach until the number of "features" does not exceed some limit - my gut feeling is ten (10).
Below this number we can live with current solution without much problems.
If, and only if, this becomes very hard to maintain and document (over 10) we can consider building such an Feature Gate solution.

pbochynski:
What about this pull request? Can we move it forward?

agolawska:
please add the decision parameters according to the decision template https://github.com/kyma-project/community/blob/master/.github/ISSUE_TEMPLATE/decision-record.md

Issue #115, opened 2018-10-24
Add the Installation doc type to the content strategy document


kazydek:
**Description**

Update the [content strategy](https://github.com/kyma-project/community/blob/master/guidelines/content-guidelines/content-strategy.md#documentation-types) document to include the **Installation** document type as an additional, optional, document type. Describe the type and its purpose in short.

**Reasons**

Since the number of Kyma installation guides is (and will be) growing we would like to expose those documents more and separate them from other Getting Started guides or Details docs on the installation topic. Thus, an idea for a new type was proposed and agreed on as part of [that](https://app.zenhub.com/workspace/o/kyma-project/kyma/issues/1352) issue.

PR #117, opened 2018-10-30
Add the installation document type

kazydek:
**Description**

- Added the **Installation** document type to the `content-strategy.md` document.

**Related issue(s)**
Relates to #115

### DIFF
--- a/guidelines/content-guidelines/content-strategy.md
+++ b/guidelines/content-guidelines/content-strategy.md
@@ -39,21 +39,30 @@ Every independent Kyma component is a separate documentation topic. The only exc
 
 ## Documentation types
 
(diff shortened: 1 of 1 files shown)


guidelines/content-guidelines/content-strategy.md
```diff
@@ -39,22 +39,31 @@ Every independent Kyma component is a separate documentation topic. The only exc
 
 ## Documentation types
 
+There is a set of documents that a given technical topic must include. You can also add additional document types to expose a specific topic better. However, before you create an additional document type, consult it with other Kyma content developers.
```
klaudiagrz: Maybe mentioning a SIG Content would be a good idea? We could also provide a link to SIG Content README so that sb who'd like to introduce changes could see when our meetings take place.

kazydek: Already discussed and clarified that.

PR #176, opened 2019-01-03
Generic Asset Store proposal

derberg:
Description

- Proposal for generic object store solution
- New diagrams and yaml files for proposal

It is a prerequisite for https://github.com/kyma-project/community/pull/107


sig-and-wg/sig-core/proposals/objects-store-proposal.md
```diff
@@ -0,0 +1,155 @@
...
+  
+![](assets/storage.svg)
+
+1. Location of the object is specified in the Object custom resource.
+2. ObjectStore controller fetches the object basing on the information given in the custom resource.
+3. The controller performs:
+    - Mutation of the object by communicating with mutation webhook specified in the custom resource
+    - Validation of the object by communicating with mutation webhook specified in the custom resource
+    - New file creation, if such file was referenced in the resource definition as a ConfigMap
+   If any of above operations failed, controller updates the resource with `ready: False` status
```
Abd4llA: This line doesn't render well and shows up as a continuation of the previous bullet point.

sig-and-wg/sig-core/proposals/objects-store-proposal.md
```diff
@@ -0,0 +1,155 @@
...
+3. Storage for static client-side applications
+
+## Solution
+  
+![](assets/storage.svg)
+
+1. Location of the object is specified in the Object custom resource.
+2. ObjectStore controller fetches the object basing on the information given in the custom resource.
+3. The controller performs:
+    - Mutation of the object by communicating with mutation webhook specified in the custom resource
```
Abd4llA: My understanding is that a mutating webhook works by registering the mutating webhook per CRD and k8s calls the webhook on CRD events directly. I believe the controller does not communicate with the webhook.

sig-and-wg/sig-core/proposals/objects-store-proposal.md
```diff
@@ -0,0 +1,155 @@
+# Objects Store
```
Abd4llA: I suggest to rename it to `ObjectStore Orchestrator`

sig-and-wg/sig-core/proposals/objects-store-proposal.md
```diff
@@ -0,0 +1,155 @@
...
+- Minio doesn't support setting a CDN for your objects. Future controller of the Bucket custom resource will have to handle this additional functionality. For example controller will be reesponsible for configuring a CloudFront for your bucket on S3
+
+Example resource for first version of the ObjectStore:
+```
+apiVersion: bucket.objectstore.kyma-project.io/v1alpha1
+kind: Bucket
+metadata:
+  name: my-bucket
+  namespace: stage
+#ObjectStore 2.0
```
Abd4llA: Are these lines commented?

sig-and-wg/sig-core/proposals/objects-store-proposal.md
```diff
@@ -0,0 +1,155 @@
...
+          replace: $2<title>$1</title>
+  filesFrom:
+    - configMapRef:
+        name: additional-object
+    - configMapRef:
+        name: one-more-object
+  bucketRef:
+    name: my-bucket
+status:
+  ready: True
```
Abd4llA: what will the status be if one of the files in the index was not fetched? what will happen with files that were uploaded so far? will they be deleted? or partially available?

sig-and-wg/sig-core/proposals/objects-store-proposal.md
```diff
@@ -0,0 +1,155 @@
...
+  objects:
+    - url: https://some.storage.domain/01-overview.md
```
Abd4llA: How will the user know which files from the index.yaml map to which objects? based on metadata title and type? does that mean title and type should be unique? 

An easier alternative is to specify the exposed base URI under which the files will be exposed in a path structure similar to the index.yaml. 
e.g.

```
  baseURI: https://some.storage.domain/
  objects:
    - url: https://some.storage.domain/01-overview.md
      metadata:
        title: MyOverview
        type: Overview
    - url: https://some.storage.domain/02-details.md
      metadata:
        title: MyDetails
        type: Details
```

Abd4llA:
Also I strongly suggesting adding a spec documentation similar to the following 
https://github.com/knative/eventing/blob/master/docs/spec/spec.md

derberg:
@Abd4llA thank a lot for your feedback, I will successively answer your question while working on completing the proposal.

Regarding general comment I prefer we use our official template for documenting the CR:
https://github.com/kyma-project/community/blob/master/guidelines/templates/resources/custom-resource.md and so far plan is that once we start implementation such final document will be provided

derberg:
@Abd4llA regarding other general question (I don't like that in github you cannot answer them under the question but as a separate comment)

- rook is rather for orchestration and set up of different storage offerings. This proposal goes further, it is not only about setting up a storage but also setting up buckets and fetching resources and the most important part, enabling mutation and validation on them. This is not in scope of rook. Also their focus is on minio and other solutions and here in Kyma we do not need this orchestration. Minio will not be/should not be used in Kyma on production but only for local dev, for production we will advice to start minio as a gateway to some other object stores like GCS, Azure Blob or Amazon S3. It will be explained in this proposal
- I thought about different scenarios and so far the best is to have just the namespaced only resources. In future we could extenda that but so far all frontend or docs related use cases are for a given namespace, you would not like to mix in one bucket content from different namespaces like production and stage
- do you really thing RBAC should be addressed by this proposal? at the end it is all based on CRDs and it is later up to the administrator to decide how to set them up. For now I would not make any specific defaults here as at the end doesn't matter who is the provider, the object store costs are really small.

sig-and-wg/sig-core/proposals/objects-store-proposal.md
```diff
@@ -0,0 +1,155 @@
...
+1. Storage for documentation and related images
+2. Storege for API specifications
+3. Storage for static client-side applications
+
+## Solution
+  
+![](assets/storage.svg)
+
+1. Location of the object is specified in the Object custom resource.
+2. ObjectStore controller fetches the object basing on the information given in the custom resource.
```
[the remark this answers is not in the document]
derberg: yeap, I'll make sure naming is consistent everywhere once I'm done, the diagram already looks a bit different in my drafts

sig-and-wg/sig-core/proposals/objects-store-proposal.md
derberg: fixed

sig-and-wg/sig-core/proposals/objects-store-proposal.md
derberg: it is not about the MutatingWebhook I think you are referring to (https://github.com/morvencao/kube-mutating-webhook-tutorial/blob/master/deployment/mutatingwebhook.yaml) as this one enables you to do a mutation of the CRD. Here we need a custom approach here, where we can enable mutation of the fetched object before it is pushed to the bucket, to enable things like rewriting, unziping and things like that. You will see more explanation once I'm done

sig-and-wg/sig-core/proposals/objects-store-proposal.md
derberg: I'm not sure we can really call it orchestrator, but maybe it is just me thinking that orchestrator is something big :)

maybe you could elaborate more

sig-and-wg/sig-core/proposals/objects-store-proposal.md
derberg: yes, because I do not see them in first implementation as those extra things need more investigation and research to really decide. I leave it there to somehow refer in the example to the first bullet point about the future, this is why I call it `ObjectStore 2.0`

sig-and-wg/sig-core/proposals/objects-store-proposal.md
derberg: the uniqueness/mapping is achieved on `01-overview.md` and `02-details.md` parts of the URL. Filenames are not modified. Metadata can be anything, no uniqueness there.

I like your baseURI idea, but then should it look something like:
```
baseURI: https://some.storage.domain/
  objects:
    - name: 01-overview.md
       metadata:
         title: MyOverview
         type: Overview
    - name: 02-details.md
       metadata:
         title: MyDetails
         type: Details
```

sig-and-wg/sig-core/proposals/objects-store-proposal.md
```diff
@@ -0,0 +1,238 @@
...
+   If any of above operations failed, controller updates the resource with `ready: False` status  
+4. Controller uploads the object to minio to a bucket that name is specified in the custom resource. You need a bucket to upload objects, you create it separately as a Bucket custom resource
+5. Controller updates the status of the Object custom resource with information about location of the file
+
+### Bucket custom resource
+
+You might want to use different bucket per solution. This is why you need to be able to specify multiple buckets for different Object. For example one bucket for documentation and one bucket per Web application.
+
+Another use cases for having multiple Buckets configuration:
+- Future extensibility by alowing bucket policy specification per solution
```
jesusreal: allowing ;)

sig-and-wg/sig-core/proposals/objects-store-proposal.md
derberg: through direct feedback I also get suggestion to remove `Object` from the name as it is to generic and also kind of hard to position and explain with Object Storage name in general.

Suggested name is AssetStore and looking on use cases I think I ❤️ it

derberg: will wait with renaming though as I'm interested what others mean

sig-and-wg/sig-core/proposals/asset-store-proposal.md
```diff
@@ -0,0 +1,274 @@
...
+## Motivation
+
+1. Have a generic solution for storing in s3 like storage any kind of asset: zip file, markdown file, PNG or JS (JavaScript)
+2. Have a storage solution not locked to one vendor, like AWS S3 only, or like to Minio on-premise only
+3. Have a storage solution that not only stores the file but also exposes it directly to internet
+
+## Use case
+
+1. Storage for documentation and related images
+2. Storege for API specifications
```
akucharska: Storage

sig-and-wg/sig-core/proposals/asset-store-proposal.md
```diff
@@ -0,0 +1,274 @@
...
+spec:
+  source:
+    mode: single # or index or package
+    url: https://some.domain.com/main.js
+  bucketRef:
+    name: my-bucket
+```
+
+The optional information is the:
+- ConfigMap reference, that points to the ConfigMap that introduces [a new file](https://github.com/kyma-project/kyma/blob/master/resources/core/charts/service-catalog-addons/charts/instances-ui/templates/configmap.yaml) that is also sent to the bucket along with other files
```
kazydek:
```suggestion
- A ConfigMap reference that points to the ConfigMap that introduces [a new file](https://github.com/kyma-project/kyma/blob/master/resources/core/charts/service-catalog-addons/charts/instances-ui/templates/configmap.yaml) which is also sent to the bucket along with other files.
```

sig-and-wg/sig-core/proposals/asset-store-proposal.md
```diff
@@ -0,0 +1,274 @@
...
+- Asset validation webhook reference to a service that performs the validation of fetched assets before they are uploaded to the bucket. The use cases are:
```
kazydek:
```suggestion
- An Asset validation webhook reference to a service that performs the validation of fetched assets before they are uploaded to the bucket. The use cases are:
```

sig-and-wg/sig-core/proposals/asset-store-proposal.md
```diff
@@ -0,0 +1,274 @@
...
+  - validation of specific file against some specification
```
kazydek:
```suggestion
  - Validation of a specific file against some specification.
```

sig-and-wg/sig-core/proposals/asset-store-proposal.md
```diff
@@ -0,0 +1,274 @@
...
+  - security validation
```
kazydek:
```suggestion
  - Security validation
```

sig-and-wg/sig-core/proposals/asset-store-proposal.md
```diff
@@ -0,0 +1,274 @@
...
+      - keyvalue: 
+          basePath: /test/v2
+  bucketRef:
+    name: my-bucket
+status:
+  phase: Ready # Failed/Pending
+  reason: Validation failure
+  message: "file is not valid against provided json schema"
+```
+- Asset mutation webhook reference to a service that acts similar to the validation service mentioned above, but mutates the asset instead of just validating. The use case is for example the asset rewriting through `regex` operation or `keyvalue` (like for example modification in json specification)
```
kazydek:
```suggestion
- An Asset mutation webhook reference to a service that acts similarly to the validation service mentioned above. The difference is that it mutates the asset instead of just validating it. The use case is, for example, the asset rewriting through the `regex` operation or `keyvalue`, or modification in the json specification).
```

sig-and-wg/sig-core/proposals/asset-store-proposal.md
```diff
@@ -0,0 +1,274 @@
...
+
+
+## Asset Validation Webhook details
```
kazydek:
```suggestion
## Asset validation webhook details
```

sig-and-wg/sig-core/proposals/asset-store-proposal.md
```diff
@@ -0,0 +1,274 @@
...
+
+Asset Store must provide a flexible way of validating the asset before they are uploaded to the bucket. Different use cases bring different validation requirements. The best way is to do it by introducing a webhook that will receive an asset for validation and reply to the controller with defined status schema
```
kazydek:
```suggestion
The Asset Store must provide a flexible way of validating the asset before they are uploaded to the bucket. Different use cases bring different validation requirements. The best way is to do it by introducing a webhook that will receive an asset for validation and reply to the controller with a defined status schema.
```

sig-and-wg/sig-core/proposals/asset-store-proposal.md
```diff
@@ -0,0 +1,274 @@
...
+
+1. Controller calls the `/validate` endpoint of the given service in a given namespace
```
kazydek:
```suggestion
1. The controller calls the `/validate` endpoint of the given service in a given Namespace.
```

sig-and-wg/sig-core/proposals/asset-store-proposal.md
```diff
@@ -0,0 +1,274 @@
...
+```
+validationWebhookService:
+  name: swagger-operations-svc
+  namespace: default
+  metadata:
+    pattern: \json|yaml
+  endpoint: "/validate"
+```
+2. The service gets the following payload and by default must reply within 1sec:
```
kazydek:
```suggestion
2. The service gets the following payload and, by default, it must reply within 1 second:
```

sig-and-wg/sig-core/proposals/asset-store-proposal.md
```diff
@@ -0,0 +1,274 @@
...
+    metadata:
+      pattern: \json|yaml
+    assets: {
+      name1: "content",
+      name2: "content"
+    }
+}
+```
+3. The controller:
+   - times out after 1sec because of no response and updates the status of the Asset to False
```
kazydek:
```suggestion
   - Times out after 1 second because of no response, and updates the status of the Asset to `False`.
```

sig-and-wg/sig-core/proposals/asset-store-proposal.md
```diff
@@ -0,0 +1,274 @@
...
+   - gets response with successful or failing validation in the following for:
```
kazydek:
```suggestion
   - Gets the response with successful or failing validation in the following form:
```

sig-and-wg/sig-core/proposals/asset-store-proposal.md
```diff
@@ -0,0 +1,274 @@
...
+       message: "much more details of the failure"
+     },
+     name2: {
+       status: Success,
+       message: "much more details"
+     }
+   }
+}
+```
+4. If at least one asset failed the validation the status of the Asset is set to False and proper message about a given asset is added to the status. Otherwise the status is set to True and asset uplodaded to the bucket.
```
kazydek:
```suggestion
4. If at least one asset failed the validation, the status of the Asset is set to `False` and the proper message about a given asset is added to the status. Otherwise, the status is set to `True` and asset is uploaded to the bucket.
```

sig-and-wg/sig-core/proposals/asset-store-proposal.md
```diff
@@ -0,0 +1,274 @@
...
+
+
+## Asset mutation webhook details
+
+Asset Store must provide a flexible way of mutating the asset before they are uploaded to the bucket. Different use cases bring different mutation requirements. Like in case of validation it can be solved by webhook solution.
```
kazydek:
```suggestion
The Asset Store must provide a flexible way of mutating the asset before they are uploaded to the bucket. Different use cases bring different mutation requirements. Similarly to the validation case, it can be solved by a webhook solution.
```

sig-and-wg/sig-core/proposals/asset-store-proposal.md
```diff
@@ -0,0 +1,274 @@
...
+
+1. Controller calls the `/mutate` endpoint of the given service in a given namespace
```
kazydek:
```suggestion
1. The controller calls the `/mutate` endpoint of the given service in a given Namespace.
```

PR #183, opened 2019-01-15
Release process improvements proposal

crabtree:
Created on 2019-02-08 by Krystian Cieślik (@crabtree).

## Decision log

| Name | Description |
|-----------------------|------------------------------------------------------------------------------------|
| Title | Release process improvements to mitigate not tested set of components |
| Ultimate decision maker(s) | Kyma Council |
| Due date | 2019-02-28 |
| Status | Proposed on 2019-02-08. |
| Input provider(s) | Kyma community |
| Group(s) affected by the decision | Kyma developers |
| Decision type | Binary |
| Earliest date to revisit the decision | 2019-06-01 |
| Affected decisions | None |

## Context

Proposal for release process improvements

## Decision

The decision is to do the implementation of improvements in the release process to minimize the risks related to releasing not tests components.

## Consequences

More stable and quicker release process.

### DIFF
--- a/sig-and-wg/sig-core/proposals/release-process-improvement.md
+++ b/sig-and-wg/sig-core/proposals/release-process-improvement.md
@@ -36,7 +36,7 @@ Having such assumptions in mind we can mitigate almost all identified problems. 
 
 #### Charts repository
 
-Following structure represents repository for charts. Under this repository, there will be two main directories: `modules` and `kyma`. In `modules` directory, there are all charts that make up the kyma. In the `kyma` directory, there is Dockerfile that is used to create the kyma-installer and two `*.yaml` files which provide deployment details for the kyma-installer.
+Following structure represents repository for charts. Under this repository, there will be two main directories: `modules` and `kyma`. In `modules` directory, there are all charts that make up the kyma. In the `kyma` directory, there is Dockerfile that is used to create the kyma-installer, two `*.yaml` files which provide deployment details for the kyma-installer and all installation related scripts.
 
 ```
 kyma
--- a/sig-and-wg/sig-core/proposals/release-process-improvement.md
+++ b/sig-and-wg/sig-core/proposals/release-process-improvement.md
@@ -1,0 +1,101 @@
+# Release process improvements
+
+## Current situation:
+
+- Components as source code getting versioned independently from kyma resources folder.
+- Kyma resources folder copied into installer identifies a kyma version (represented by kyma-installer docker image).
+- The release process is taking all the resources and building a kyma version with the lates versions from the components source code.
+
+## Problems:
+
+- The release job consists of two factors - the pipeline definition (described in cont-integration repository) and the pipeline itself defined in Jenkinsfile under the component directory in the consolidated kyma repository. The definition of the pipeline may be removed or renamed as the component evolves. It is hard to determine the particular state of the cont-integration repository from which the release was performed. It may happen that when we remove a component and then remove its job definition the particular bugfix release will no longer be possible. Migration to prow somehow mitigates this problem as the test-infra repository will also have a release branch defined. Unfortunately, jobs configuration is still defined on the master branch of test-infra repository so no job definitions should be deleted until we are sure that we will not need it for a bugfix release.
+- The current master branch version is based on components in a specific version which is not necessarily the latest. Docker images set in the charts not necessarily point to the latest builds of the components. When creating a release branch, suddenly the potential release will consist of all components in the latest version. That is a mismatch which is unexpected and will cause problems. Maybe people by purpose have not updated the component version in the charts. The solution would be either we assure that always the latest version is used (which we tried with the dynamic versioning but people have not accepted it) or we do not change the used versions for a release, which is the current proposal.
+- Each run of release pipeline requires building of all the components - running unit tests, compiling the source code, building and pushing the docker images. Unfortunately, we do not utilize the caching mechanism offered by the docker on our building nodes (jenkins slaves, prow job containers).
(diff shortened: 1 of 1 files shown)


sig-and-wg/sig-core/proposals/release-process-improvement.md
```diff
@@ -0,0 +1,107 @@
...
+
+In this proposal I identify two release artifacts which are:
+- Helm charts of all components which then compose our product
+- Basic yaml files for kyma-installer (config yaml referencing the installer image, containing the charts, referencing the components)
+
+The source code of the components is not a release artifact, but it must be possible and easy to track a particular commit in the repository from which the docker image was produced.
+
+We should split the kyma repository into two. First containing charts and installation resources, second the source code (components/tools/tests)
+
+Having such assumptions in mind we can mitigate almost all identified problems. Focusing the release on the charts repository we are sure that what we release is what we already tested. We do not have to freeze the source code repository for a release as the process do not rely on it. All docker images of the components are already released and tested, we do not need to rerun the building process.
```
pbochynski: I don't understand why do you need to separate charts or modules to mitigate the problem. The only change which you propose is NOT releasing components (code) when we release kyma. As I understand kyma version will contain charts that will have docker images of kyma components with tagged with commitId. If you want to fix it you need to check out component with this commitId (repository doesn't matter - can be the same), fix the bug, build new image with new commitId and update chart.

In my opinion it is not a step in the right direction. We keep all components isolated and we give up continuous integeration. I see the vision differently. I would like to build all the components with every pull request and have them tested continuously all the time (now we test them together only after docker image bump). I would start the proposal from the vision - how we imagine building kyma in 6 months. For me continuous integration brings more value than independent development of components.

sig-and-wg/sig-core/proposals/release-process-improvement.md
```diff
@@ -0,0 +1,107 @@
...
+- The current master branch version is based on components in a specific version which is not necessarily the latest. Docker images set in the charts not necessarily point to the latest builds of the components. When creating a release branch, suddenly the potential release will consist of all components in the latest version. That is a mismatch which is unexpected and will cause problems. Maybe people by purpose have not updated the component version in the charts. The solution would be either we assure that always the latest version is used (which we tried with the dynamic versioning but people have not accepted it) or we do not change the used versions for a release, which is the current proposal.
+- Each run of release pipeline requires building of all the components - running unit tests, compiling the source code, building and pushing the docker images. Unfortunately, we do not utilize the caching mechanism offered by the docker on our building nodes (jenkins slaves, prow job containers).
```
pbochynski: In the last release it took about 12 minutes., and I think we can further optimize it. Is it a real problem?

sig-and-wg/sig-core/proposals/release-process-improvement.md
sjanota: The main reason I can think of is that it would be misleading to have in the same repository code and charts that are not related to each other. Moreover, as @crabtree pointed out we need branch per docker image to provide fixes. It will create a mess in `kyma` repository and I think we should strive to have it as readable as possible. 

In my opinion we have two repositories merged into one, but we never do changes in both repositories at once. It is reflected even in our workflow: you have to create two PRs to make change in a component.

pbochynski: Ok. I understand now why you want to split repositories. We will have nice repo with charts (with release branches), and not so nice repo with components where you need to know the commitId to get a code for given component. 
I think we have different visions how kyma project should look like in the future. 
The vision I derive from your proposal is about:
- reposiotory per module/area
- independent releases (for each repo)
- kyma release is the chart with references to different versions of modules (we don't release all modules with kyma releases)
- integration testing done on kyma chart level (change tested when component version is upgraded in chart)
My vision I can summarize:
- monorepo with all kyma components,
- more common patterns, libraries (maybe common vendor)
- run integration tests continuously with every commit to source code

Both visions have pros and cons. 

My solution requires more resources in the build system (always build all components) and requires use of feature toggles to not break builds when implementing new features. But you can have release branches that contain a code which is released. It also simplifies cross project changes (upgrade kubernetes version or kubernetes client version and be done in one pull request).

Your solution is more efficient for build/release infrastructure. But cross project initiatives will be harder to execute and it will be a challenge to achieve consistency between modules. Different release cycles for different components can also impact our productivity.

I would like to have a section at the beginning of the proposal that contains the vision/direction we want to drive in kyma-project.

crabtree: @aszecowka I agree that my proposal is intrusive and requires changes in the process. Can you elaborate a bit more about your proposed solution? As this is a periodic job, I understand that this will be only some kind of check, that indicates whether components latest in particular point in time can work together. Or as periodic you mean some kind of nightly build? How that PR would be then merged into the master branch?

aszecowka: @crabtree This periodic job will check if the code in the master is "mergeable".
- creates a PR as a bot "kyma-bumper". This PR contains changes to values.yaml's for components that were changed recently
- integration tests for PR are executed. If everything is fine, we need to gather approvals from code reviewers and merge PR. This needs to be done by Kyma developer so we need somehow appoint such person responsible for that. Later we can think also about automation and implement automatic merges.

agolawska:
please add the decision parameters according to the decision template https://github.com/kyma-project/community/blob/master/.github/ISSUE_TEMPLATE/decision-record.md

crabtree:
As we have half a year left from the initial proposal and it becomes somehow outdated I decided to close this PR.

PR #185, opened 2019-01-15
Prepare proposal with common validator for Kyma

michal-hudy:
**Description**

- Proposal with common validator for Kyma

**Related issue(s)**

Resolves kyma-project/kyma#1844

### DIFF
--- a/sig-and-wg/sig-core/proposals/common-validator.md
+++ b/sig-and-wg/sig-core/proposals/common-validator.md
@@ -354,7 +354,7 @@ $(eval $(call SANDBOX,lint))
 
 And `Makefile` in component:
 ```makefile
-APP_NAMES = name_1 name_2 name_3
+APP_NAMES = name1 name2 name3 name4
 BUILDPACK_VERSION = v20181119-afd3fbd
 REPOSITORY_PATH = $(realpath $(shell pwd)/../..)
 
@@ -367,7 +367,7 @@ $(foreach appName,$(APP_NAMES),$(eval $(call TARGETS,$(appName))))
 
 To unify our validation and `Makefile` targets we should define a template that will be included in all components `Makefiles`. Thanks to that we will be able to have one place per repository with targets definitions and it will be easier to introduce new validations.
 
-Such proposal also introduces the possibility to validate source code against CI environment, what can speed up creating Pull Requests. It is possible to support multiple artifacts per components, but we should consider if we will still want it.
+Such proposal also introduces the possibility to validate source code against CI environment, what can speed up creating Pull Requests. It is possible to support multiple artifacts per components, but we should consider if we still want it.
 
 Examples:
  - [One artifact per component](#sandbox-validation)
--- a/sig-and-wg/sig-core/proposals/common-validator.md
+++ b/sig-and-wg/sig-core/proposals/common-validator.md
@@ -212,9 +212,9 @@ include $(REPOSITORY_PATH)/scripts/template.go.mk
 $(eval $(call TARGETS,main.go,Dockerfile))
 ```
 
-Thanks to `template.mk` we will have one place in repository with targets definitions and it will be much easier to maintain it.
+Thanks to `template.go.mk` we will have one place in repository with targets definitions and it will be much easier to maintain it.
 
->**NOTE**: All components needs to be build by CI system in case of changes in `template.mk`.
+>**NOTE**: All components needs to be build by CI system in case of changes in `template.go.mk`.
 
 ### Sandbox Validation
 
--- a/sig-and-wg/sig-core/proposals/common-validator.md
+++ b/sig-and-wg/sig-core/proposals/common-validator.md
@@ -354,7 +354,7 @@ $(eval $(call SANDBOX,lint))
 
 And `Makefile` in component:
 ```makefile
-APP_NAMES = ui-api-layer
+APP_NAMES = name_1 name_2 name_3
 BUILDPACK_VERSION = v20181119-afd3fbd
 REPOSITORY_PATH = $(realpath $(shell pwd)/../..)
 
--- a/sig-and-wg/sig-core/proposals/common-validator.md
+++ b/sig-and-wg/sig-core/proposals/common-validator.md
@@ -8,7 +8,7 @@ Proposed on 2019-01-15.
 
 ## Motivation
 
-The Kyma project is a one big mono repository with multiple separated repositories. That repositories are validated in a different ways. Current situation causes following issues:
+The Kyma project is a one big mono repository with multiple separate projects. That projects are validated in a different ways. Current situation causes following issues:
 
 - Script that validates repositories is duplicated across components.
 - Multiple versions of scripts that validates components.
@@ -58,11 +58,13 @@ As a solution for that I would like to unify targets:
  - `validate` - execute all validation
  - `resolve` - download dependencies
  - `build` - execute build of component
+ - `clean` - remove artifacts
  - `test` - execute tests for component
  - `lint` - execute linters for component
  - `format` - execute code formating for component
- - `build-image` - build docker image with component
- - `push-image` - push docker image to repository
+ - `validate-format` - validates if files are formated correctly
+ - `docker-build` - build docker image with component
+ - `docker-push` - push docker image to repository
 
  After unification, `makefile` for `ui-api-layer` component will look like that:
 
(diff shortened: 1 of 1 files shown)


sig-and-wg/sig-core/proposals/common-validator.md
```diff
@@ -0,0 +1,266 @@
...
+push-image:
+	docker tag $(APP_NAME) $(IMG):$(TAG)
+	docker push $(IMG):$(TAG)
+ ```
+
+### Common Makefile
+
+As in previous point unified targets were introduced it would be fine to not duplicate all that `makefile` content in other components. Fortunately it is possible, because it is possible to include `makefiles`. Thanks to that we can create one common `makefile` that will be included in components `makefiles`.
+
+Lets name this common `makefile` `template.mk` - files with `*.mk` are threated as `makefiles` - and put it in `kyma/common` directory.
```
sjanota: I'm not a fan of placing it there. `kyma/comon` is a regular go library and has very different purpose then storing templates. I'd suggest placing it at root level.

sig-and-wg/sig-core/proposals/common-validator.md
```diff
@@ -0,0 +1,266 @@
...
+
+### Sandbox Validation
+
+Developers often have different environment than CI system. For example different version of Golang. It is important to provide a way for verifying code against CI environment. For that lets create another target `validate-sandbox`, that will execute all verification in same `buildpack` Docker image as CI system is using.
+
+Definition of `validate-sandbox` target in `template.mk`:
+```makefile
+COMPONENT_REL_PATH=$(shell echo $(shell pwd) | sed 's,$(REPOSITORY_PATH)/,,g')
+
+validate-sandbox:
```
sjanota: 👍 I suggest adding sandboxed versions of other commands too. For example `gofmt` works differently in go 1.10 and go 1.11. As a developer I don't need only to validate if my code is working with go 1.10, but I need also to fix it using tools from go 1.10.

sig-and-wg/sig-core/proposals/common-validator.md
```diff
@@ -0,0 +1,266 @@
...
+IMG = $(DOCKER_PUSH_REPOSITORY)$(DOCKER_PUSH_DIRECTORY)/$(APP_NAME)
+TAG = $(DOCKER_TAG)
+
+.PHONY: ci-pr ci-master ci-release resolve build test format lint build-image push-image
+
+ci-pr: validate build-image push-image
+ci-master: ci-pr
+ci-release: ci-master
+
+validate: resolve build test format lint
```
sjanota: Why do we even need build in here? Isn't test enough?

sig-and-wg/sig-core/proposals/common-validator.md
```diff
@@ -0,0 +1,266 @@
...
+
+define TARGETS
+resolve:
+	dep ensure -v -vendor-only
+build:
+	go build -o $(APP_NAME) $(1)
+test:
+	go test ./...
```
crabtree: We may also provide `-race` flag to `go test` to enable race detection.

sig-and-wg/sig-core/proposals/common-validator.md
```diff
@@ -0,0 +1,266 @@
...
+format:
+	go fmt
+lint:
+	go vet
+endef
+
+build-image:
```
marcobebway: In general the proposal looks good to me. 

I Just want to mention that in the event-bus component, we are building/pushing three docker images, so we have to override the `build-image` and `push-image` recipes in the event-bus makefile. Will this cause any issues or concerns?

michal-hudy: I'm aware of that, and I will update `template.mk` to handle such situation. I will ping you when I update it 😄

derberg: @marcobebway please keep in mind that there is also a separate proposal from @a-thaler that is already accepted by many people that assumes split of event-bus directory. So anyway it should not be a problem. Would be actually an awesome stuff in for 0.8 scope we could go forward with implementation of this proposal and also cleanup of directories structure/name

sig-and-wg/sig-core/proposals/common-validator.md
michal-hudy: Good point, `kyma/common` was just an example location. I think it shouldn't be in the `root` of the repository, because we may want to introduce also some common scripts for validating or something in the future.

sjanota: I would also rename the file to `template.go.mk` as we have other technologies also here, like bash scripts etc. They are not something we are proud of, but we cannot just ignore our children ;)

sig-and-wg/sig-core/proposals/common-validator.md
```diff
@@ -0,0 +1,266 @@
...
+Developers often have different environment than CI system. For example different version of Golang. It is important to provide a way for verifying code against CI environment. For that lets create another target `validate-sandbox`, that will execute all verification in same `buildpack` Docker image as CI system is using.
+
+Definition of `validate-sandbox` target in `template.mk`:
+```makefile
+COMPONENT_REL_PATH=$(shell echo $(shell pwd) | sed 's,$(REPOSITORY_PATH)/,,g')
+
+validate-sandbox:
+	docker run --rm -v "$(REPOSITORY_PATH):/workspace/go/src/github.com/kyma-project/kyma" \
+	--workdir "/workspace/go/src/github.com/kyma-project/kyma" \
+	eu.gcr.io/kyma-project/prow/test-infra/buildpack-golang:v20181119-afd3fbd \
```
sjanota: I have one problem with sandbox version: version of buildpack is duplicated (here and in CI) and may diverge. In such case it will be misleading. Also it assumes all components uses same buildpack which is not necessarily true. Do we care and if so how do we want to solve it?

sig-and-wg/sig-core/proposals/common-validator.md
```diff
@@ -0,0 +1,266 @@
...
+`ui-api-layer`:
+```makefile
+APP_NAME = ui-api-layer
+REPOSITORY_PATH = $(realpath $(shell pwd)/../..)
+include $(REPOSITORY_PATH)/common//template.mk
+
+$(eval $(call TARGETS,main.go))
+```
+
+Thanks to `template.mk` we will have one place in repository with targets definitions and it will be much easier to maintain it.
```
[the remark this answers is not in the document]
sjanota: Until [proposal for splitting kyma repo](https://github.com/kyma-project/community/pull/183) is approved it's not really disadvantage as we have only one repository.

sig-and-wg/sig-core/proposals/common-validator.md
michal-hudy: The main focus of this proposal is architecture, not what is executed in `make` targets. I expect also discussion about `gometalinter`, `goimports` or other tools, but it should be discussed during implementation I think.

Anyway, I will add this flag to the proposal.

sig-and-wg/sig-core/proposals/common-validator.md
```diff
@@ -0,0 +1,266 @@
...
+
+define TARGETS
+resolve:
+	dep ensure -v -vendor-only
+build:
+	go build -o $(APP_NAME) $(1)
+test:
+	go test ./...
+format:
+	go fmt
```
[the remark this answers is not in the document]
michal-hudy: Please see https://github.com/kyma-project/community/pull/185#discussion_r248251233

sig-and-wg/sig-core/proposals/common-validator.md
michal-hudy: Yes, because we want to validate if the build works ;)

sig-and-wg/sig-core/proposals/common-validator.md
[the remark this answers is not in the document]
michal-hudy: Sure, I will add a note. I don't see a way to fix it now as we cannot expect that developer will also clone some other repository with common scripts etc. and also this should work offline.

sig-and-wg/sig-core/proposals/common-validator.md
michal-hudy: I will add an environment variable to `makefile` on component level with `buildpack` version. Unfortunately, it will be still duplicated in `makefile` and CI job definition. If we will start using new tags for `buildpacks` then it will look better. By new tag, I mean `go1.11` that I introduced with new `buildpack` 😄

sig-and-wg/sig-core/proposals/common-validator.md
michal-hudy: @marcobebway Proposal updated, but as @derberg mentioned, we should consider if we still want it 😄

sjanota:
I wonder if it makes sense to split architecture and content of targets. We cannot start with implementation without both of those things.

mszostok:
@sjanota IMO yes, this is just the first phase. First, let's unify the whole thing in all components using what we already have. Then we can have a second proposal about the quality itself saying what kind of check should be executed. There we can debate  if we should use the -race flag, if we should have a hammer-test with flag `-count 1000`, if we should use [gometalinter](https://github.com/alecthomas/gometalinter) or [golangci-lint](https://github.com/golangci/golangci-lint) and so on and so forth. Why do we want to save the whole world in one proposal?

sjanota:
@mszostok Sounds good to me. Just reminder that we shouldn't jump straight to coding without all parts in place.

PR #350, opened 2019-08-27
Proposal for upgrades of Kyma on latest K8s

derberg:
**Description**

- Proposal for upgrades of Kyma on latest K8s

**Resolves**

this PR resolves kyma-project/kyma/issues/4853

### DIFF
--- a/collaboration/sig-core/proposals/regular-k8s-upgrades.md
+++ b/collaboration/sig-core/proposals/regular-k8s-upgrades.md
@@ -1,0 +1,31 @@
+# Regular upgrades of Kyma to work on latest Kubernetes
+
+Created on 2019-08-27 by Michal Hudy, Marcin Witalis and Lukasz Gornicki.
+
+## Motivation
+
+We should regularly validate Kyma compatibility with latest K8s version to not create technology dept and adjust and react properly to comming changes.
+
+Examples of the issue:
+- some time ago apiVersions of some workloads, like Deployments were set to be deprecated. We could use new versions seamlesly few months ago. Unfortunately it didn't happen and because copy/paste we used many deprecated areas of K8s that causes a lot of issues for Kyma upgrades. 
+- we know that we are compatible with 1.12 K8s while the latest version is 1.15. This is not the best situation that accumulates possible upgradability problems
+- we do not know if Kyma works on **kind** and latest **minikube**
+
+## Solution
+
+We need to have in place the following:
+* CI
+  * Weekly CI jobs that runs Kyma installation and upgrade on latest [kind](https://github.com/kubernetes-sigs/kind) (kind is the recommended CI tool for testing K8s)
+  * Weekly CI jobs that runs Kyma installation and upgrade on latest minikube
+* Every release we need a person responsible to:
+  * Update the jobs to point to latest kind, minikube and k8s
+  * Review release notes and new features presentation and prepare a summary for all Kyma maintainers on important features and deprecations
+  * Review the results of the jobs that 
+    * fail and creates GitHub issues to provide fixes. If the same fixes are required in several components, this person should provide recommendation how to provide a fix.
+    * are successful and provides updates to documentation about support K8s version
+  * Checks AKS and GKE support for K8s and makes sure that in docs and CI we always use latest. Managed offerings are always behind. For example now K8s 1.15 is available but managed offerings support 1.13 as latest, with 1.14 as beta.
+
+## Action items
+
+* Create CI plans. Work on **kind** is delayed by Service Catalog that in Kyma still requires separate etcd. Because 1.16 is not yet released, we can work on pipeline for Kyma 1.7 timeline in October
+* Create a clear procedure/guideline how to perform manual work so the whole upgrade flow is simple to go through by any maintainer. 


tehcyx:
1) Long-term I'm interested if **kind** will replace **minikube**, so we can reduce the costs again.

2) Is this regarding every Kyma release or regarding every release of **kind**, **minikube** and **k8s**?
> Every release we need a person responsible to:
>   * Update the jobs to point to latest **kind**, **minikube** and **k8s**

3) What are we going to do with old k8s releases? Is there a minimum version requirement that we want to support? You're speaking of latest k8s release support and that we should test this. Does this mean with every k8s release we just add one more job but don't remove old jobs? When do we stop supporting older version?

derberg:
@tehcyx thanks for taking a look. Let me answer one-by-one
- I think it should be natural that `kind` replaces `minikube` for testing. We still need to make sure Kyma runs on `minikube`, but regular testing should be done on `kind` imho. This is not something that requires a proposal, afaik Prow by default supports `kind`, so once we make sure it works and start implementing this proposal, you can already start integrating it PR jobs
- it is about every release of kind, minikube and k8s. So this will be regularly done and for now, I'm the volunteer here with folks around Core and Supporting capability, but we want to document it properly so anyone can support here
- yes, this proposal is only for the latest releases and assumption is, no new jobs, but always update of the last job to use the latest kind/minikube/k8s. Your question is still valid. For this Kyma release, 1.5, we did manual testing against k8s 1.15 and on minikube and could not do clusters as GKE and AKS supported 1.13 only. So far Kyma release strategy says, use latest Kyma, and same should be with K8s in my opinion. We should always run our pipelines against latest k8s and simply recommend to use latest as keeping pipelines for few versions back will generate costs. Do you think this is related to this proposal?

tehcyx:
@derberg glad to hear about point 1. and 2.

To your last point, I think it depends on this proposal. We see already that we can't test against the latest GKE, AWS, ... we can only deploy on vanilla k8s, testing it manually. We will only find out e.g. for the 1.5 release, that if it doesn't work on a GKE 1.15 cluster once that is available. Until then we can only safely say it is tested on 1.15 vanilla k8s. But until 1.15 is available we still need to run it on 1.13 GKE. Which already gives us two different k8s versions to test. On the other hand I can create clusters in GKE right now that range from 1.11 to 1.13. Should I be able to install Kyma on all of them? Does that connect with the marketplace offering we have e.g. on the GKE Marketplace or can we map Marketplace releases to GKE cluster versions and make them unavailable in case we say 1.13 is our minimum supported version?

derberg:
@tehcyx this requires rethinking the strategy for Kyma availability. Afaik if something works on `kind` it works everywhere, so maybe we will be able to test on `kind` only? Of course, managed offerings are still a problem here. Without testing on older version I cannot say that Kyma always will work there as I don't know if someone maybe used a feature that is not available in older versions. 

any thoughts from your side here. Always test the latest GKE and AKS and latest on `kind` and this way we can kind of reach a wider spectrum of versions...thinking process

tehcyx:
I'm fine with test on latest hyperscaler, but does running on **kind** really mean that it runs everywhere? E.g. afaik it does currently not run on **kind** but works on GKE, AKS and minikube. To be fair I also only tried the 1.4.x release on kind locally yesterday. We'd have to create different jobs that use different kind cluster configurations as well, e.g.  `kind create cluster --image kindest/node:v1.12.10@sha256:e43003c6714cc5a9ba7cf1137df3a3b52ada5c3f2c77f8c94a4d73c82b64f6f3` for a 1.12 cluster.

I don't want to start rethinking the strategy in this issue here, but for me the proposal entails a lot more than just `adding support for kind`.

derberg:
@tehcyx proposal is not only about kind, it is about more, as you can see we want to also test against latest minikube as it is still our recommended way of using kyma locally for example. Also proposal is about addressing future releases, analysis of release notes and spreading the knowledge.
This proposal is only about checking Kyma against latest and addressing issues. The things you raise is also important though

Btw, I extended deadline for feedback for a proposal for a week

collaboration/sig-core/proposals/regular-k8s-upgrades.md
```diff
@@ -0,0 +1,31 @@
...
+
+We need to have in place the following:
+* CI
+  * Weekly CI jobs that runs Kyma installation and upgrade on latest [kind](https://github.com/kubernetes-sigs/kind) (kind is the recommended CI tool for testing K8s)
+  * Weekly CI jobs that runs Kyma installation and upgrade on latest minikube
+* Every release we need a person responsible to:
+  * Update the jobs to point to latest kind, minikube and k8s
+  * Review release notes and new features presentation and prepare a summary for all Kyma maintainers on important features and deprecations
+  * Review the results of the jobs that 
+    * fail and creates GitHub issues to provide fixes. If the same fixes are required in several components, this person should provide recommendation how to provide a fix.
```
mmitoraj:
```suggestion
    * if any jobs fail, the person should create GitHub issues to provide fixes. If the same fixes are required in several components, this person should provide recommendation on how to provide a fix.
```

collaboration/sig-core/proposals/regular-k8s-upgrades.md
```diff
@@ -0,0 +1,31 @@
...
+    * are successful and provides updates to documentation about support K8s version
```
mmitoraj:
```suggestion
    * if the jobs are successful, the person should provide updates to documentation about the support of a particular K8s version
```

collaboration/sig-core/proposals/regular-k8s-upgrades.md
```diff
@@ -0,0 +1,31 @@
...
+  * Checks AKS and GKE support for K8s and makes sure that in docs and CI we always use latest. Managed offerings are always behind. For example now K8s 1.15 is available but managed offerings support 1.13 as latest, with 1.14 as beta.
```
mmitoraj:
```suggestion
  * checking AKS and GKE support for K8s and makes sure that in docs and CI we always use the latest. Managed offerings are always behind. For example, now K8s 1.15 is available but managed offerings support 1.13 as the latest, with 1.14 as beta.
```

collaboration/sig-core/proposals/regular-k8s-upgrades.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Action items
+
+* Create CI plans. Work on **kind** is delayed by Service Catalog that in Kyma still requires separate etcd. Because 1.16 is not yet released, we can work on pipeline for Kyma 1.7 timeline in October
```
mmitoraj:
```suggestion
* Create CI plans. Working on **kind** is delayed because of the Service Catalog that in Kyma still requires separate etcd. Because 1.16 is not released yet, we can work on a pipeline for Kyma 1.7 timeline in October.
```

Issue #360, opened 2019-09-18
Update guidelines with info on using `...` to signalize ellipsis in code


majakurcius:
**Description**

Update content guidelines and add info to use `...` in code blocks in docs to signalize that part of code has been omitted.

Use:
```
{
  ...
  "certificate":{
    "subject":"O=Organization,OU=OrgUnit,L=Waldorf,ST=Waldorf,C=DE,CN=CNAME",
    "extensions":"",
    "key-algorithm":"rsa2048"
  }
}
```

**Reasons**

It's not always necessary to quote code in full, e.g. full HTTP responses. Often, it's only a small part that's of interest to the user. For the sake of brevity, skip the irrelevant part. For the sake of clarity, replace it with `...` to let the user know that it's not the whole thing.

PR #361, opened 2019-09-20
Update guidelines with info on ellipsis in code

majakurcius:
**Description**

- Update content guidelines and add info to use `...` in code blocks in docs to signalize that part of code has been omitted.

**Related issue(s)**
See #360

### DIFF
--- a/guidelines/content-guidelines/04-formatting.md
+++ b/guidelines/content-guidelines/04-formatting.md
@@ -60,7 +60,7 @@ For example, you might have an HTTP response as such:
 }
 ```
 
-Suppose, that it is only the part containing the certificate information that is of interest to the user. In that case, instead of quoting the whole reponse, quote it like this:
+Suppose that it is only the part containing the certificate information that is of interest to the user. In that case, instead of quoting the whole response, quote it like this:
 ```json
 {
   ...
--- a/guidelines/content-guidelines/04-formatting.md
+++ b/guidelines/content-guidelines/04-formatting.md
@@ -39,7 +39,7 @@ Metadata names            | When you create a Markdown document, define its `tit
 
 ### Omission in code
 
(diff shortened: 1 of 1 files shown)


guidelines/content-guidelines/04-formatting.md
```diff
@@ -36,6 +36,42 @@ Metadata names            | When you create a Markdown document, define its `tit
 >**NOTE:** When you mention specific configuration files in your documents, consider linking to them instead of just mentioning their names. When you link to a file, use its name without the format extension. See the following example:
 > `To adjust the number of Pods in your Deployment, edit the [deployment](./deployment.yaml) file.`
 
+
+### Omission in code
+
+Often, when you quote snippets of code, it is not necessary to include them in full. For example, when you quote an HTTP response, most often, it is enough to only quote the relevant part, and there is no need to enclose the whole thing. In such a case, replace the omitted parts with `...` to signalize to the reader that something has been removed.
```
klaudiagrz:
```suggestion
Often, when you quote snippets of code, it is not necessary to include them in full. For example, when you quote an HTTP response, it is enough to quote only the relevant part instead of enclosing the whole response. In such a case, replace the omitted parts with `...` to signalize that something has been removed. 
```

guidelines/content-guidelines/04-formatting.md
```diff
@@ -36,6 +36,42 @@ Metadata names            | When you create a Markdown document, define its `tit
...
+  },
+  "certificate":{
+    "subject":"O=Organization,OU=OrgUnit,L=Waldorf,ST=Waldorf,C=DE,CN=CNAME",
+    "extensions":"",
+    "key-algorithm":"rsa2048"
+  }
+}
+```
+
+Suppose, that it is only the part containing the certificate information that is of interest to the user. In that case, instead of quoting the whole reponse, quote it like this:
```
klaudiagrz:
```suggestion
Suppose that it is only the part containing the certificate information that is of interest to the user. In that case, instead of quoting the whole response, quote it like this:
```

PR #383, opened 2019-10-11
Proposal for consolidating Kyma controllers inside a single controller manager

antoineco:
**Description**

- Consolidate Kyma controllers inside a single controller manager

**Related issue(s)**

kyma-project/kyma#4327

### DIFF
--- a/collaboration/sig-core/proposals/kyma-controller-manager.md
+++ b/collaboration/sig-core/proposals/kyma-controller-manager.md
@@ -1,0 +1,181 @@
+# Consolidate Kyma controllers inside a single controller manager
+
+Created on 2019-10-11 by Antoine Cotten (@antoineco).
+
+## Status
+
+DRAFT. Proposed on 2019-10-11.
+
+## Motivation
+
+Kyma is composed of multiple components that share a lot of architectural patterns, yet are written in isolation
+following different engineering standards and using heterogeneous SDKs and libraries, before eventually being released
+as _one product_.
+
+Most of these components involve one or more Kubernetes controllers that synchronize API objects using identical
+building blocks: watches, work queues, reconciliation loops, etc. Every controller instance maintains its own local
+cache of several Kubernetes resources, which can result in a significant memory footprint in productive environments.
+Besides, each of these controllers is operated in a similar fashion by nature: as a single instance or in an
+active-passive setup.
+
+These similarities represent an opportunity for us to consolidate Kyma controllers and run them under the umbrella of a
+single _manager_, so that they can reuse common libraries, maintain a consistent level of compatibility with the
+underlying Kubernetes layer, share memory efficiently, and avoid the fragmentation we are currently observing in the
+project.
+
+This mode of operation is directly inspired by the Kubernetes [controller-manager][k8s-cm].
+
+## Goals
+
+This initiative aims at improving three aspects:
+
+**Maintainability**
+
+Streamline the codebase regarding controllers to harmonize the different implementation aspects. (_"Common opinion"_)
+
+**Footprint**
+
+Reduce the resource consumption by sharing internal API object caches whenever possible. (_"Store once"_)
+
+**Operability**
+
+Improve operational aspects by reducing the number of controllers that have to be deployed separately. (_"Run once"_)
+
+## Constraints
+
+It must be possible to opt out of each individual controller inside the manager to keep Kyma modular. This could be
+achieved using startup flags and/or a ConfigMap.
+
+A controller ("operator") SDK should be agreed upon by all code owners. [kyma#4327][kyma4327] initiated
+a discussion around that topic, but no conclusion has been reached at the time of writing. For the initial version of
+this document, we assume that [`kubebuilder`][kb] will be our SDK of choice. A few reasons to believe this assumption is
+safe to make:
+- well maintained and documented by a dedicated group (SIG) within the Kubernetes project
+- leverages `controller-runtime`, a collection of sensible libraries with everything a controller typically needs
+- bootstaps projects with the manager pattern enabled right from the start
+- includes tooling oriented towards future extensibility, with helpers to generate new APIs
+- RBAC objects are part of the code generation
+
+## Proposed solution
+
+### Project structure
+
+The structure of the controller manager is initialized from _scratch_, existing Kyma components get incorporated
+gradually into it after suitable candidates have been identified.
+
+A rough draft of what the structure of the Kyma controller manager could look like:
+
+```
+├── api
+│   ├── applicationconnector
+│   │   └── v1alpha1
+│   │       ├── application_types.go
+│   │       ├── <other types>…
+│   │       ├── groupversion_info.go
+│   │       └── zz_generated.deepcopy.go
+│   ├── eventing
+│   │   └── v1alpha1
+│   │       ├── subscription_types.go
+│   │       ├── <other types>…
+│   │       ├── groupversion_info.go
+│   │       └── zz_generated.deepcopy.go
+│   ├── serverless
+│   │   └── v1alpha1
+│   │       ├── function_types.go
+│   │       ├── <other types>…
+│   │       ├── groupversion_info.go
+│   │       └── zz_generated.deepcopy.go
+│   └── <other API groups>…
+│
+├── controllers
+│   ├── applicationconnector
+│   │   ├── application_controller.go
+│   │   ├── <other resource controllers>…
+│   │   └── pkg
+│   ├── eventing
+│   │   ├── subscription_controller.go
+│   │   ├── <other resource controllers>…
+│   │   └── pkg
+│   ├── serverless
+│   │   ├── function_controller.go
+│   │   ├── <other resource controllers>…
+│   │   └── pkg
+│   └── commons
+│
+├── Dockerfile
+├── go.mod
+├── go.sum
+├── main.go
+└── Makefile
+```
+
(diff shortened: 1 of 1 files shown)


collaboration/sig-core/proposals/kyma-controller-manager.md
```diff
@@ -0,0 +1,181 @@
...
+- bootstaps projects with the manager pattern enabled right from the start
+- includes tooling oriented towards future extensibility, with helpers to generate new APIs
+- RBAC objects are part of the code generation
+
+## Proposed solution
+
+### Project structure
+
+The structure of the controller manager is initialized from _scratch_, existing Kyma components get incorporated
+gradually into it after suitable candidates have been identified.
```
ormos: Who would own that project the codebase for that controller manager? Right now the different controllers are owned by individual teams, who are managing the lifecycle of that component. How are we planning to coordinate that if we have a central controller manager?

collaboration/sig-core/proposals/kyma-controller-manager.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Status
+
+DRAFT. Proposed on 2019-10-11.
+
+## Motivation
+
+Kyma is composed of multiple components that share a lot of architectural patterns, yet are written in isolation
+following different engineering standards and using heterogeneous SDKs and libraries, before eventually being released
+as _one product_.
```
ormos: I fully agree and that is also something which we should not only drive forward regarding the controllers. There are a lot of other more generic technical aspects (e.g. logging) which we should harmonize.

collaboration/sig-core/proposals/kyma-controller-manager.md
antoineco: Totally agree. The teams will still own their APIs and controllers, but we need some governance to ensure all code owners are on the same page when it comes to project structure, dependencies management, etc.

collaboration/sig-core/proposals/kyma-controller-manager.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Constraints
+
+It must be possible to opt out of each individual controller inside the manager to keep Kyma modular. This could be
+achieved using startup flags and/or a ConfigMap.
+
+A controller ("operator") SDK should be agreed upon by all code owners. [kyma#4327][kyma4327] initiated
+a discussion around that topic, but no conclusion has been reached at the time of writing. For the initial version of
+this document, we assume that [`kubebuilder`][kb] will be our SDK of choice. A few reasons to believe this assumption is
+safe to make:
```
mszostok: here is my comment about going with the kubebuilder:
https://github.com/kyma-project/kyma/issues/4327#issuecomment-541596179

Some time ago I advocated the kubebuilder approach. Also, some of Gopher components are written with kubebuilder. But after some time I've changed my opinion about that and I've described that in the linked issue.

collaboration/sig-core/proposals/kyma-controller-manager.md
```diff
@@ -0,0 +1,181 @@
...
+- leverages `controller-runtime`, a collection of sensible libraries with everything a controller typically needs
+- bootstaps projects with the manager pattern enabled right from the start
+- includes tooling oriented towards future extensibility, with helpers to generate new APIs
+- RBAC objects are part of the code generation
+
+## Proposed solution
+
+### Project structure
+
+The structure of the controller manager is initialized from _scratch_, existing Kyma components get incorporated
```
mszostok: We need to also start the business discussion here. Because IMO base on that this proposal can evaluate in two different directions:

**One direction:**
Go with one controller manager + give an option to extract single controller in future

**Second direction:**
Go with domain controller managers as we have right now e.g. asset store. In this case, need more like the best practices for writing the controller.

Because I want to know if this proposal is:
-  really about having only one controller-manager in Kyma. 
- OR best practices, agreements for writing the controller (libs, pkg layout, logging, scaling etc)

antoineco: This proposal is about having a single controller manager to fulfill the 3 goals defined in the document. Defining best practices for actually _writing_ the controllers, is not incompatible with that, but it was not my main intention.

collaboration/sig-core/proposals/kyma-controller-manager.md
antoineco: I agree with your analysis. I'm don't agree with all of kubebuilder's opinions myself, we have to strike a good balance and be ready to fill some gaps on our own. I just didn't want to send a proposal that suggests to agree on a SDK but doesn't take any position, because the discussion would be harder to start without an initial assumption.

collaboration/sig-core/proposals/kyma-controller-manager.md
mszostok: So IMO we should upfront define which components will be there:
- can we move all of them into on-part (I know that we have components like Application Broker which consists of OSB Api Spec Broker and controller, same for Helm Broker and how to deal with that?),
- how much controllers we have in Kyma (i.e. I know that we recently extracted some of them into separate repos like helm-broker, asset-store, api-gateway, etc.), 
- and how we want to group them, only one manager, or we require to have more like domain managers 

so thanks to that we will have a clear fishion what to do after accepting this proposal, i.e. which controllers will be grouped together, if this proposal also touches controller extracted directly to kyma-project, and so on and so forth

mszostok:
@derberg this is exactly what I wrote 15days ago: https://github.com/kyma-project/community/pull/383#discussion_r334802258

collaboration/sig-core/proposals/kyma-controller-manager.md
a-thaler: Modularity is an important aspect of Kyma and there are scenarios were people deselect features. So either that single controller manager activates the different controller aspects dynamically dependent on the current installation setup, or we really have a manager per domain. I think that decision can be done more easy seeing the actual set of kyma controllers.

piotrmiskiewicz: Some components used in Kyma are separate components like Helm-Broker. Such component won't be a part of the main controller manager.  
As I see, we see the following controllers (maybe I missed something):
 * api-controller
 * application-broker (exposes HTTP endpoints)
 * application-operator
 * asset-store (will be extracted to its own repository)
 * cms (will be extracted to its own repository)
 * connection-token-handler
 * connectivity-certs-controller
 * function-controller
 * service-binding-usage-controller
 * kyma-operator

Other components are independent applications and won't be a part of the new controller manager. As @mszostok wrote, the application-broker contains an HTTP server which expose OSB API.

From the testing point of view we need an easy way to run a particular controller.

aszecowka:
> Every controller instance maintains its own local cache of several Kubernetes resources, which can result in a significant memory footprint in productive environments.

Can you provide some real data?

Also, I am a little bit afraid that one poor controler can impact other controllers, for example memory or resource leak will cause that all our controllers will be restarted constantly.
I am also wondering about security aspects, right now every controller has limited access to k8s objects he can read/modify via service account, role and role binding. If we are going to have only one deployment, then how we want to ensure that for example Service Catalog CRDs are modified only by given controller?

collaboration/sig-core/proposals/kyma-controller-manager.md
strekm: At some point, api-controller will be replaced with api-gateway, which is already extracted from kyma. kyma-operator is special case controller which installs/updates kyma itself and in my opinion either it should be extracted from kyma or shouldn't be part of discussed here controller-manager

antoineco:
> Also, I am a little bit afraid that one poor controler can impact other controllers, for example memory or resource leak will cause that all our controllers will be restarted constantly.

This reasoning is dangerous and will harm the project long term. It's the duty of ALL maintainers to ensure Kyma respects some standards of quality and that all controllers are thoroughly tested. "I want to write my test in isolation because others write poor code" is exactly what we should avoid.

> Can you provide some real data?

No, but I believe the concept is obvious. Imagine a cluster with 500 common objects like Pods, Services, ... With an average of 10KB per object in memory (shouldn't be too far from reality), the cache of a single controller is 5MB just for those objects. Multiply by ~15 controllers, and add the more specialized objects controllers may have in cache (Deployments, StatefulSets, Knative objects...). 10MB may not be high, but the sum of those caches is not negligble.

> If we are going to have only one deployment, then how we want to ensure that for example Service Catalog CRDs are modified only by given controller?

You're right, we won't get that with a single controller.

collaboration/sig-core/proposals/kyma-controller-manager.md
sjanota: A real-life scenario where kubebuilder has bitten us: one of CBS dependencies was forcing kubernetes 1.15, but kubebuilder was using older controller-runtime at that time (pre 2.0 times). We ended up with a feature postponed until kubebuilder 2.0

antoineco: At that point it's clear kubebuilder is a bad choice for us, so this implementing this proposal will require a decision for kyma-project/kyma#4327.

derberg:
I've heard we are close to a deadline for a review but also current questions not addressed. What is the deadline for? after deadline the questions will be answered or?

collaboration/sig-core/proposals/kyma-controller-manager.md
antoineco: Removed references to kubebuilder.

PR #386, opened 2019-10-14
Add initial proposal of introducing Knative Eventing Mesh

Abd4llA:
**Description**

- Add a proposal to introduce Knative Brokers instead of Event Bus

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/5926


sig-and-wg/sig-core/proposals/introducing_knative_brokers.md
```diff
@@ -0,0 +1,88 @@
...
+## Features
+
+### Current Features
+
+- As a developer, I want to publish events to Kyma from outside the cluster.
+- As a developer, I want to publish events to Kyma from inside the cluster.
+- As a developer, I want to subscribe to events in Kyma from a specific source with a specific event-type.
+- As an admin, I want to enable or disable events in a specific namespace.
+
+### New features
```
sayanh: This is not a new feature. Right now, the cm `default-ch-webhook` exists which enables this functionality.

sig-and-wg/sig-core/proposals/introducing_knative_brokers.md
```diff
@@ -0,0 +1,90 @@
...
+
+- As an admin, I want to configure the default cluster wide Knative `Channel` kind to be used. 
+- As an admin, I want to configure a Knative `Channel` kind to be used by default for events per namespace that overrides the cluster default configuration.
+
+## Proposed solution
+
+There are two proposed solutions. Solution 1 is an instant transition and deprecation proposal where the new `Broker` based eventing will replace the existing Event Bus. The Kyma `Susbscription` and `EventActivation` CRDs will be removed and Event Bus users will need to migrate to the new knative CRDs.
```
sayanh: - `new Broker/Trigger` based
- `EventActivation` CRDs will not be removed now until we solve the Event registry workflow.

sig-and-wg/sig-core/proposals/introducing_knative_brokers.md
```diff
@@ -0,0 +1,90 @@
...
+2. For every new namespace, Kyma Console ensures it has the label `knative-eventing-injection=enabled` so that a default `Broker` gets created by the Knative Eventing controller.
+3. Once the `Application Broker` gets a provision request for a `Service Instance` which has `bindable` parameter set to `false`, it will create a `Knative Subscription` that has:
+   - The channel pointing to the relevant `Event Service` fanout `Channel`
+   - The subscriber pointing to the default `Broker` of the namespace
+   - The ownerRef set as the `Service Instance` that just got provisioned which ensures the subscription gets deleted by Kubernetes garbage collector when the `Service Instance` gets deprovisioned.
+4. `Kyma Console` creates the proper `Trigger` objects for the `Lambda` function for every event type chosen as a lambda trigger.
+5. Events received by `Event Service` will be delivered to its channel which in turn will do a fanout to all connected brokers which will finally deliverd it to the lambda functions.
+
```
sayanh: I think it would make sense to add the deprovisioning process in a different section. We need to answer:
- Who deletes the Knative Subscription to the fanout channel? IMO the `application` should be the owner of the Knative Channel.

In the Knative Brokers approach, it is not clear how are we solving this:
> - As a developer, I want to publish events to Kyma from inside the cluster.

sig-and-wg/sig-core/proposals/introducing_knative_brokers.md
```diff
@@ -0,0 +1,90 @@
+# Integration of the Knative-based function controller
+
+## Introduction
+
+As of 0.5, Knative Eventing defines [`Broker` and `Trigger`](https://knative.dev/docs/eventing/broker-trigger/) objects as higher level abstractions that allows brokering events based on some event filtering criterea.
```
sayanh:
```suggestion
As of 0.5, Knative Eventing defines [`Broker` and `Trigger`](https://knative.dev/docs/eventing/broker-trigger/) objects as higher-level abstractions that allow brokering events based on some event filtering criteria.
```

sig-and-wg/sig-core/proposals/introducing_knative_brokers.md
```diff
@@ -0,0 +1,90 @@
...
+
+This is a proposal of how to introduce these objects in Kyma as an alternative to using the existing Kyma Event Bus. It describes how to build namespace scoped eventing mesh with fault domain isotation and access control. The proposal describes two possible solutions, the migration path and tries to highlight and address the challenges and open questions of using them.
```
sayanh:
```suggestion
This is a proposal of how to introduce these objects in Kyma as an alternative to using the existing Kyma Event Bus. It describes how to build namespace scoped eventing mesh with fault domain isolation and access control. The proposal describes two possible solutions, the migration path and tries to highlight and address the challenges and open questions of using them.
```

sig-and-wg/sig-core/proposals/assets/knative_brokers_arch.svg
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
sayanh: typo: `provisions`

sig-and-wg/sig-core/proposals/introducing_knative_brokers.md
```diff
@@ -0,0 +1,90 @@
...
+1. For every new `Application` the `Application Operator` creates an `Event Service` and its corresponding  `Knative Channel` that will be used to fanout events to the opted-in namespaces.
+2. From Kyma Console UI, when the user creates a namespace, the pop up will show an info message and toggle with something like:
+>Kyma now comes with a new `Knative Broker` based eventing feature (beta).
+>Would you like to try it out for this namespace?. Read more about it here <link to more docs>
+and by enabling the toggle, Kyma Console ensures it has the label `knative-eventing-injection=enabled` so that a default `Broker` gets created by the Knative Eventing controller.
+3. Once the `Application Broker` gets a provision request for a `Service Instance` which has `bindable` parameter set to `false`, it creates a `Knative Subscription` for `opted-in` namespaces (as in solution 1) or an `EventActivation` for `legacy namespaces`.
+4. `Kyma Console` creates the appropriate `Trigger` or `Kyma Subscription` objects for the `Lambda` function for `opted-in` and `legacy` namespaces.
+5. Publishing events from `Publish` will deliver events to subscribers in both `opted-in` and `legacy` namespaces.
```
sayanh: I have a few concerns with proposal 2:
- Publish service now needs to be a controller and needs to keep a state of all the new knative channels(for all applications) created to fanout the events. This is something I would avoid.
- What will be the response when one of the publish fails? E.g. there are 2 channels for 2 applications and a channel(old one). Will we fail the publish request when 1 channel fails and rest succeed? What are the implications?
- There should be another migration plan for this approach to a purely new approach which should be here IMO.
- We are not eliminating Publish Service but adding more logic to it which is against our objective.

sig-and-wg/sig-core/proposals/introducing_knative_brokers.md
```diff
@@ -0,0 +1,90 @@
...
+### New features
+
+- As an admin, I want to configure the default cluster wide Knative `Channel` kind to be used. 
+- As an admin, I want to configure a Knative `Channel` kind to be used by default for events per namespace that overrides the cluster default configuration.
+
+## Proposed solution
+
+There are two proposed solutions. Solution 1 is an instant transition and deprecation proposal where the new `Broker` based eventing will replace the existing Event Bus. The Kyma `Susbscription` and `EventActivation` CRDs will be removed and Event Bus users will need to migrate to the new knative CRDs.
+
+Solution 2 is about having both systems coexisting for one or two releases, during which the namespaces with the `knative-injection-enabled=true` label will have the new `Broker` based eventing, while the ones that don't will just have the same old Kyma Event Bus based behavior (i.e uses Kyma `EventActivation` and `Subscription` CRDs) in order to:
```
sayanh:
```suggestion
Solution 2 is about having both systems coexisting for one or two releases, during which the namespaces with the `knative-eventing-injection=enabled` label will have the new `Broker` based eventing, while the ones that don't will just have the same old Kyma Event Bus based behavior (i.e uses Kyma `EventActivation` and `Subscription` CRDs) in order to:
```
Let's be consistent with the label.

abbi-gaurav:
I just had a discussion with @sayanh and I see the following concerns. Please let me know if the mechanism works differently
With the proposed model, we will end up sending all the events from the application to the broker and there it will get filtered. 
This implies that even if we have a lambda trigger set for one event, we will still incur the network, storage and compute the cost for all the events that are sent from the application. This cost will either manifest with the customer or whoever is managing the runtime.
Moreover, with all the network and resource noise created by unwanted events, the delivery and performance for interesting events will be impacted.

Such a problem will manifest when a customer has a peak volume such as black Fridays and could end up being charged multiple times than intended due to the channel-subscription design.

I believe since this is a new redesign, we can possibly alleviate all such concerns at the start.
Not sure, but in the past, there were discussions about cluster level broker triggers. 

@Abd4llA, we can discuss it further in the office.

sig-and-wg/sig-core/proposals/introducing_knative_brokers.md
```diff
@@ -0,0 +1,89 @@
...
+2. For every new namespace, Kyma Console ensures it has the label `knative-eventing-injection=enabled` so that a default `Broker` gets created by the Knative Eventing controller.
+3. Once the `Application Broker` gets a provision request for a `Service Instance` which has `bindable` parameter set to `false`, it will create a `Knative Subscription` that has:
+   - The channel pointing to the relevant `Event Service` fanout `Channel`
+   - The subscriber pointing to the default `Broker` of the namespace
+   - The ownerRef set as the `Service Instance` that just got provisioned which ensures the subscription gets deleted by Kubernetes garbage collector when the `Service Instance` gets deprovisioned.
+4. `Kyma Console` creates the proper `Trigger` objects for the `Lambda` function for every event type chosen as a lambda trigger.
+5. Events received by `Event Service` will be delivered to its channel which in turn will do a fanout to all connected brokers which will finally deliverd it to the lambda functions.
```
vaikas-google: nit 'delivered' (the second one is deliverd)

davyodom:
I agree with the above comments that option 1 seems preferable in general, assuming that an automatic upgrade path is supported.  I have some concerns about the implementation though, sorry if they've been addressed already.

Creating a separate broker for each namespace with a fanout channel seems like it might be an undesirable approach for Kyma.  This leads to a lot of additional complexity, internal compute and traffic cost, and many fold replication of events in the underlying messaging infrastructure (1 time for each namespace + 1 for the channel in the kyma-integration namespace).  As Gaurav points out above, the previous behavior of discarding events in the kyma 'event source' (aka Event Service) before writing to the messaging infrastructure when there are no active subscriptions seems like something that should remain as well. We don't want to put the burden on the external applications to have to worry about enabling event types one by one, they should ideally be agnostic of how their events are being consumed in extensions.  If the cluster incurs a big additional cost just by connecting applications even if nothing is consuming those events or only a small subset is being consumed, that would put some roadblocks in front of adoption.

Also, is it necessary to support different broker types in a first pass?  It seems like this wouldn't be a feature many users would leverage.  They might eventually get more sophisticated and request such things (lowered qos to decrease cost or increase performance for example),  but that seems like a long way off.  I see the desire to connect different event sources being much more likely in the near to mid term.

Abd4llA:
Thank you all for your feedback. I just wanted to provide a quick update on the proposal for everybody.

* We are currently working on updating the proposal with a new "desired state" proposal based on some of the feedback from you and also based on our findings regarding some gaps in Knative Broker.
* We will use the final proposal as an initial point to drive issues into Knative through. 
* Once we push the updated proposal, we will ask for another round of feedback

sig-and-wg/sig-core/proposals/introducing_knative_brokers.md
```diff
@@ -0,0 +1,89 @@
...
+
+Solution 1 describes how Kyma system will build and wire the eventing mesh to achieve proper event routing from external application to the subscribers such as the lambda functions.
+
+#### Architecture
+
+For every namespace that is allowed to receive events, there is a default `Broker` in the namespace. Kyma will do all the wiring from the `Event Service` to the local namespace `Brokers` and subscribers who want to receive events, should create a Knative `Trigger` against their local default `Broker`.
+
+![architecture](assets/knative_brokers_arch.svg)
+
+1. For every new `Application` the `Application Operator` creates an `Event Service` and its corresponding  `Knative Channel` that will be used to fanout events to all allowed brokers.
```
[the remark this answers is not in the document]
sayanh: >  if you use the SubscriberSpec.Ref so you would have to use the URI instead 
Exactly, we are planning to use the URI here.
```
apiVersion: eventing.knative.dev/v1alpha1
kind: Subscription
metadata:
  name: broker-sub-test
  namespace: kyma-integration
spec:
  channel:
    apiVersion: messaging.knative.dev/v1alpha1
    kind: Channel
    name: test-app-ch
  reply: {}
  subscriber:
    uri: http://default-broker.<target-namespace>/
```

Abd4llA:
We have started in Knative discussing how to provide a more efficient broker to broker event routing that would allow us a near-to-source events filtering. An idea is to introduce a broker bridge between two brokers A -> B, that allows broker A to do efficient event forwarding only if broker B has corresponding triggers. We will drive this topic through this issue knative/eventing#2050

ormos:
We have already customers, partners and early adopters being live with Kyma or planning to go live with Kyma in the next couple of weeks. So the only way forward can be option 2. With option 1 Kyma would be a big risk for these customers and partners because we can not guarantee the same functionality and behavior as the current Kyma releases. We first need to get to some experience using the EventBroker approach in customer scenarios with real-world loads before we can fully switch.

I know that make the current implementation at some level more complex but risking our first users and early adopters would be a bigger risk.

sig-and-wg/sig-core/proposals/introducing_knative_brokers.md
```diff
@@ -0,0 +1,89 @@
+# Integration of the Knative-based function controller
```
marcobebway: The title should be about introducing Knative brokers in Kyma.

a-thaler:
@Abd4llA What is the status of the proposal?

anishj0shi:
The New Event- Mesh is already live with kyma 1.11.

Current Design is based on event-sources, so every *Application* which is paired with  kyma is treated as an event source of type *HTTPSource* .

Having said that, upgrade from kyma 1.10 --> 1.11 should not break your system as we have carefully attached a migration job to ensure the migration from event-bus to event-mesh is a smooth experience. 
For more details on the event-mesh, feel free to refer our docs: https://kyma-project.io/docs/components/knative-eventing-mesh/#overview-overview

PR #485, opened 2020-07-08
Fix labels in emeritus.md

bszwarc:
**Description**

- Fixed labels


bszwarc:
That is indeed a large blob.

PR #522, opened 2020-12-02
Removing the Kyma Meetup from website

valentinvieriu:
**Description**

Removing the Kyma Meetup from website


majakurcius (approved):
image

PR #526, opened 2020-12-21
Propose of logs stream destination and library recomendation

dbadura:
**Description**

- recommend log stream destination
- recommend logging library  

**Related issue(s)**


internal/proposals/logs/streams/apex.go
```diff
@@ -0,0 +1,17 @@
...
+	"os"
+)
+
+func main() {
+	log.SetHandler(json.New(os.Stderr))
+	testApexLogger(log.WithField("request_id", "dajdhaskj"))
+}
+
+func testApexLogger(logEntry *log.Entry) {
+	logEntry.Infof("just normal log with msg: %s", "Hello From Zap")
```
franpog859, line 15: I'd rather say "Hello from Apex" ;D

franpog859 (approved):
Looks cool! 
Shouldn't we at least suggest which of these libraries we'd like to use in the components? I mean just one library so it'd be just a no-brainer for the implementing team. It'd be good to specify all the possibilities for the logging levels

internal/proposals/logs/streams/apex.go
dbadura: Good catch

internal/proposals/logs/streams/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+# Summary
+
+I would recommend logging everything to stderr, because:
+- zap, zerolog and apex/log and I think other libraries can log everything to stderr 
+- it's possible to filter logs by level and it's not needed to filter the logs by stream
+- requires less work on current components
+
+For new components, the recommendation is `Zap`.
+For already components, we should use what we currently have and unified log messages.
```
janmedrek, line 87: I would get rid of that sentence and the next one as well. As far as I know, the vast majority of libraries currently used in Kyma do not support log levels/formats/streams. We want to have the logging unified and using the common library is only making that easier.

internal/proposals/logs/streams/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+## apex/log
+pros:
+- ability to chain loggers (add context) using `entry` struct
+- library can log in JSON or text format. It has the ability to be extended.
+- configurable level of filtering
+- available log levels: ERROR, INFO, FATAL, DEBUG, WARN
+
+cons:
+- cannot set timestamp format. The default time format looks like this:  `2020-12-22T12:52:39.906885+01:00`
```
janmedrek, line 68: Doesn't it eliminate this library from the pool? That's one of our basic assumptions. If it does then please include this information here.

janmedrek (changes requested):

internal/proposals/logs/streams/README.md
dbadura: I will add test which proofs that this timestamp is valid with RFC3339 and also add explanation.

internal/proposals/logs/streams/README.md
dbadura: I added information that if library can't meet requirements should be exchanged.

janmedrek (dismissed):

kazydek (approved):
🎉

PR #533, opened 2021-01-20
Change log levels on stage and prod

piotrmiskiewicz:
**Description**

 - change the default log level for PROD and STAGE environment

Reason:
Usually we are looking at logs when something goes wrong or seems to be working wrong. Without INFO we are not able to see what was before, we do not see the history.
On the other hand there are some situations where somebody claims that our service is not working correctly. Without any logs it is hard to verify it.

If INFO logs take too much space - think about changing the level of some logs to DEBUG.


piotrmiskiewicz:
If it is ok for a team that the log level for a component could be WARN - it is ok for me to set such value, but I don't like to have a rule for all components.

PR #566, opened 2021-06-02
Final PoC Draft

JeremyHarisch:
We now had our final discussion on how the JobManager should be implemented, this is now the final sketch for it. There will be some changes during the implementation and maybe documented later on in this document.


NHingerl (approved):
for customer-facing docu, I'd prefer full stops at the end of full sentences, but let's not delay this any longer :D

PR #608, opened 2021-11-19
Merge three README.md templates into one

### LINKED ISSUE #603: Merge README.md templates into one
**Description**

Currently, we have three README templates in the `community` repo:
- [README (general)](https://github.com/kyma-project/community/blob/main/templates/repository-template/README.md)
- [README for charts](https://github.com/kyma-project/community/blob/kyma2/templates/resources/chart_README.md)
- [README for examples](https://github.com/kyma-project/community/blob/main/templates/resources/example_README.md)

Having so many README templates is confusing for the developers and contributors, and since the templates are very similar, it would be good to merge them into one general template. When working on a unified template, add a note there that not all sections are obligatory, but the order of the sections should be followed. After creating the unified template, remove the old ones and update the table with doc templates. 

AC:
- [ ] Add a new, unified template for README.md files
- [ ] Remove the old templates from [`templates/resources`](https://github.com/kyma-project/community/tree/main/templates/resources)
- [ ] Update the [templates table](https://kyma-project.io/community/guidelines/templates/templates-type/)

**Area**

area/documentation

**Reasons**

Having so many README templates is confusing and unnecessary. 

**Assignees**

@kyma-project/technical-writers

mmitoraj:
**Description**

- Merged three README.md templates into one

**Related issue(s)**
Resolves #603


templates/repository-template/README.md
```diff
@@ -1,23 +1,35 @@
...
 
 ## Prerequisites
 
-List the requirements to run the project.
+> List the requirements to run the project or example.
 
 ## Installation
 
-Explain the steps to install your project. Include the instructions or provide links to the related documentation.
+> Explain the steps to install your project.
```
NHingerl, line 19: I'd add the information that the steps should be a numbered list (or more than one, if there are several separate tasks to do). It's implied in the word "steps" but maybe readers think of subheadlines etc instead (whereas I think subheadlines should outline the structure if there are more than one set of steps).

templates/repository-template/README.md
```diff
@@ -1,23 +1,35 @@
...
+> Explain how to use the project. You can create multiple subsections (H3). Include the instructions or provide links to the related documentation.
 
 ## Development
 
-Add instructions on how to develop the project. It must be clear what to do and how to trigger the tests so that other contributors know how to make their pull requests acceptable. Include the instructions or provide links to related documentation.
+> Add instructions on how to develop the project or example. It must be clear what to do and, for example, how to trigger the tests so that other contributors know how to make their pull requests acceptable. Include the instructions or provide links to related documentation.
+
+## Troubleshooting
+
+> List potential issues and provide tips on how to avoid or solve them.
```
NHingerl, line 35: We could suggest to use the same structure as for the troubleshooting doc type - symptom, cause, remedy.

templates/repository-template/README.md
```diff
@@ -1,23 +1,35 @@
-# Project Title
+> **NOTE:** It is a general template that can be used for a project README.md, example README.md, or any other README.md type in all Kyma repositories in the Kyma organization. Not all the sections are mandatory. Use only those that suit your use case but keep the proposed section order.
```
NHingerl, line 1: if we say "not all of them are mandatory", we might want to point out specifically which ones are optional. Or are all of them optional? Then we need to rephrase the "not all are mandatory" sentence.

templates/repository-template/README.md
mmitoraj: Yeah! I will add it.

templates/repository-template/README.md
mmitoraj: OK, good point Nina :)

templates/repository-template/README.md
mmitoraj: I guess all of them are optional (maybe except for the Project Title and Overview?). I will mark that somehow :)

NHingerl: Then I'd just add "(mandatory)" behind those two :D

NHingerl (approved):

PR #636, opened 2022-03-09
Infrastructure reconciliation

pbochynski:
**Description**

- proposal for reconciliation changes to address infrastructure layer

### DIFF
--- a/concepts/infrastructure-reconciliation/README.md
+++ b/concepts/infrastructure-reconciliation/README.md
@@ -1,39 +1,41 @@
-# Motivation
-Reconciler is a framework to install, update and repair Kyma components in managed Kyma Runtimes. The reconciliation runs in the loop to ensure that all components are properly configured, up and running. This principle can be extended to cover also infrastructure components like virtual networks or kubernetes clusters, but requires some adjustments in the reconciler architecture. 
+# Reconciler infrastructure changes proposal
 
-# New use cases for reconciler
+## Motivation
+Reconciler is a framework to install, update, and repair Kyma components in managed Kyma Runtimes. The reconciliation runs in the loop to ensure that all components are up and running and properly configured. This principle can be extended to cover also infrastructure components, such as virtual networks (VNet) or Kubernetes clusters, but requires some adjustments in the Reconciler architecture. 
 
-- create VNET for Kyma Runtime 
-- create/reconcile shoot cluster for Kyma Runtime 
-- produce kubeconfig for provisioned clusters (can be rotated)
-- support bring your own cluster models (BYOC) - cluster is ready and we got cluster-admin kubeconfig 
-- trigger reconciliation on configuration or cluster state change
-- register the Kyma Runtime within the CMP
+## New use cases for Reconciler
 
-# Proposed changes
+- Create VNet for Kyma Runtime 
+- Create/reconcile a shoot cluster for Kyma Runtime 
+- Create kubeconfig for provisioned clusters (can be rotated)
+- Support bring your own cluster models (BYOC) - a cluster is ready and we get cluster-admin kubeconfig 
+- Trigger reconciliation on a configuration or on a cluster state change
+- Register Kyma Runtime within the CMP
 
-## Cluster Inventory as a separate service
+## Proposed changes
 
-Cluster Inventory is the internal component responsible for reconciliation persistence. It keeps cluster information, configuration and reconciliation status. Component reconcilers do not have access to the inventory and cannot modify it directly. One option would be to extend action context passed to the reconciler to include more data, but the better option would be to pass only information how to access cluster inventory entry (url/id) and allow components reconciler to query and update it.
+### Cluster Inventory as a separate service
 
-## Watches and triggers
-Reconciliation is currently time based (triggered in regular intervals). It is known limitation. In some cases reconciliation should be triggered immediately after some events, like:
-- configuration change (e.g. network reconciler can add vnet name to the cluster inventory what should trigger shoot reconciler)
-- resource in the cluster modified (e.g. connectivity service instance created in the cluster should trigger connectivity proxy reconciler)
+Cluster Inventory is the internal component responsible for reconciliation persistence. It keeps cluster information, configuration, and reconciliation status. Component Reconcilers do not have access to the inventory and cannot modify it directly. One option would be to extend the action context passed to the Reconciler to include more data, but the better option would be to pass only information on how to access Cluster Inventory entry (URL/ID) and allow Components Reconciler to query and update it.
 
-Restoring managed resources accidentally (or intentionally) deleted by user is not a main goal of watchers. They should be used rather to monitor resources like service instances that should trigger some integration tasks in the Kyma Control Plane.
+### Reconciliation based on watchers and triggers
+Reconciliation is currently time-based (triggered in regular intervals). It is a known limitation. In some cases, the reconciliation should be triggered immediately after some events, such as:
+- Configuration change (e.g. Network Reconciler can add a VNet name to the Cluster Inventory, which would trigger Shoot Reconciler)
+- Resource in the cluster modified (e.g. Connectivity Service instance created in the cluster would trigger Connectivity Proxy Reconciler)
 
-Watchers and triggers can be more selective to limit the number of unnecessary invocations. Component reconcilers can specify which Cluster Inventory fields changes they want to be triggered with (i.e. Istio reconciler subscribing to the kubeconfig field). 
+Restoring managed resources accidentally (or intentionally) deleted by a user is not the main goal of watchers. They should be rather used to monitor resources such as service instances that should trigger some integration tasks in the Kyma Control Plane.
 
-## No sequences or dependency management
-Reconciler tries to manage dependencies in the simple way. The prerequisites group installed first and then other components. Introducing infrastructure reconciler will make dependency management more complex. Moreover, dependencies are valid mainly for the first installation phase, and later on are not so important (in the second reconciliation eventing reconciler can be invoked before network reconciler). It should be responsibility of component reconciler to figure out if all dependencies exist and the reconciliation can be completed. If dependencies are not ready component reconciler should return the information that it is not ready and wait for the next trigger (time or event).
+Watchers and triggers can be more selective to limit the number of unnecessary invocations. Component Reconcilers can specify which changes in the Cluster Inventory fields they would be triggered by (i.e. Istio Reconciler subscribing to the kubeconfig field). 
 
-# High Level Architecture
+### No sequences or dependency management
+Reconciler tries to manage dependencies in a simple way. The prerequisites group is installed first, before other components. Introducing Infrastructure Reconciler will make dependency management more complex. Moreover, dependencies are valid mainly for the first installation phase and are not so important later on (in the second reconciliation, Eventing Reconciler can be invoked before Network Reconciler). It should be the responsibility of the Component Reconciler to figure out if all dependencies exist and the reconciliation can be completed. If dependencies are not ready, Component Reconciler should return the information that it is not ready and wait for the next trigger (time or event).
+
+## High level architecture
 
 ![](assets/infrastructure-reconciliation.svg)
 
-Mothership reconciler should provide common functionality related to watching changes and triggering component reconcilers in response to events or with time based schedule. Such aspects like queueing reconciliation requests, consolidating them, updating reconciliation status, retries and back-off strategy should be implemented once (in the mothership). 
+Mothership Reconciler should provide common functionality related to watching changes and triggering Component Reconcilers in response to events or with the time-based schedule. Such aspects as queueing reconciliation requests, consolidating them, updating reconciliation status, retries and back-off strategy should be implemented once (in the Mothership). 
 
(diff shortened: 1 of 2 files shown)


concepts/infrastructure-reconciliation/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+# High Level Architecture
+
+![](assets/infrastructure-reconciliation.svg)
+
+Mothership reconciler should provide common functionality related to watching changes and triggering component reconcilers in response to events or with time based schedule. Such aspects like queueing reconciliation requests, consolidating them, updating reconciliation status, retries and back-off strategy should be implemented once (in the mothership). 
+
+Worker pool implementation has to be revisited. Mothership should not manage dependencies - component reconciler should be written in the way that can recognize missing dependencies and can fail fast. The next reconciliation invocation should happen when the configuration (cluster inventory) is updated (dependency)
```
janmedrek, line 33: We could even think about creating some sort of subscription mechanism. For example, allow the component reconcilers to determine upon which field changes they want to be triggered (i.e. Istio reconciler subscribing to the kubeconfig field). This would reduce the load on the component reconcilers as those would not be triggered on every Cluster Inventory change, it would require some implementation effort though.

Worth giving it a thought at least, perhaps it will be a "nice to have" thing in the future, not essential for now.

concepts/infrastructure-reconciliation/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+Cluster Inventory should be extracted from the mothership and should have more flexible content. Component reconcilers should be able to read and update the content (TODO - access control).
```
janmedrek, line 35: Just a note - we also need to account for the Watchers being in different environments, so far it seems like we have KCP, SKRs, perhaps even CMP will want to plug into this flow?

concepts/infrastructure-reconciliation/README.md
```diff
@@ -0,0 +1,35 @@
+# Motivation
+Reconciler is a framework to install, update and repair Kyma components in managed Kyma Runtimes. The reconciliation runs in the loop to ensure that all components are properly configured, up and running. This principle can be extended to cover also infrastructure components like virtual networks or kubernetes clusters, but requires some adjustments in the reconciler architecture. 
+
+# New use cases for reconciler
+
+- create VNET for Kyma Runtime 
+- create/reconcile shoot cluster for Kyma Runtime 
+- produce kubeconfig for provisioned clusters (can be rotated)
+- support bring your own cluster models (BYOC) - cluster is ready and we got cluster-admin kubeconfig 
+- trigger reconciliation on configuration or cluster state change
```
janmedrek, line 10:
```suggestion
- trigger reconciliation on configuration or cluster state change
- register the Kyma Runtime within the CMP
```

janmedrek (commented):
Dropped a few comments, those are rather observations, perhaps you will want to include some of those. 🙂

Regarding the Cluster Inventory - I am all in for the "pass access information along with ID" version, this will simplify a lot of things; Mothership will no longer need to differentiate between different "types" of sub-reconcilers.

koala7659:
On your design change detection mechanism is located in the box representing Kyma cluster. Do you plan to implement is as  a separate service running inside Kyma cluster?

concepts/infrastructure-reconciliation/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+# Proposed changes
+
+## Cluster Inventory as a separate service
+
+Cluster Inventory is the internal component responsible for reconciliation persistence. It keeps cluster information, configuration and reconciliation status. Component reconcilers do not have access to the inventory and cannot modify it directly. One option would be to extend action context passed to the reconciler to include more data, but the better option would be to pass only information how to access cluster inventory entry (url/id) and allow components reconciler to query and update it.
+
+## Watches and triggers
```
akgalwas, line 18: I have one question concerning watcher in Kyma cluster. I assume that the watcher will observe some limited subset of resources to trigger reconciliation immediately in some cases (e.g. certs for Istio Gateway deleted). In most cases modifications performed by the user will be restored by time based trigger. Is that correct?

tobiscr: Plan is to watch all resources which have a particular label. The reconciler will frequently (e.g. each minute) check for identified changes and trigger then a reconciliation (we have to use pull-communication caused to security restrictions: access from SAP networks to the Internet is acceptable but not vs.).

tobiscr (approved):

concepts/infrastructure-reconciliation/README.md
pbochynski: Agree.

concepts/infrastructure-reconciliation/README.md
```diff
@@ -0,0 +1,39 @@
+# Motivation
+Reconciler is a framework to install, update and repair Kyma components in managed Kyma Runtimes. The reconciliation runs in the loop to ensure that all components are properly configured, up and running. This principle can be extended to cover also infrastructure components like virtual networks or kubernetes clusters, but requires some adjustments in the reconciler architecture. 
+
+# New use cases for reconciler
+
+- create VNET for Kyma Runtime
```
klaudiagrz, line 6:
```suggestion
- Create VNet for Kyma Runtime 
```

concepts/infrastructure-reconciliation/README.md
```diff
@@ -0,0 +1,39 @@
+# Motivation
+Reconciler is a framework to install, update and repair Kyma components in managed Kyma Runtimes. The reconciliation runs in the loop to ensure that all components are properly configured, up and running. This principle can be extended to cover also infrastructure components like virtual networks or kubernetes clusters, but requires some adjustments in the reconciler architecture. 
+
+# New use cases for reconciler
+
+- create VNET for Kyma Runtime 
+- create/reconcile shoot cluster for Kyma Runtime 
+- produce kubeconfig for provisioned clusters (can be rotated)
+- support bring your own cluster models (BYOC) - cluster is ready and we got cluster-admin kubeconfig
```
klaudiagrz, line 9:
```suggestion
- Support bring your own cluster models (BYOC) - a cluster is ready and we get cluster-admin kubeconfig 
```

concepts/infrastructure-reconciliation/README.md
```diff
@@ -0,0 +1,39 @@
...
+Reconciler is a framework to install, update and repair Kyma components in managed Kyma Runtimes. The reconciliation runs in the loop to ensure that all components are properly configured, up and running. This principle can be extended to cover also infrastructure components like virtual networks or kubernetes clusters, but requires some adjustments in the reconciler architecture. 
+
+# New use cases for reconciler
+
+- create VNET for Kyma Runtime 
+- create/reconcile shoot cluster for Kyma Runtime 
+- produce kubeconfig for provisioned clusters (can be rotated)
+- support bring your own cluster models (BYOC) - cluster is ready and we got cluster-admin kubeconfig 
+- trigger reconciliation on configuration or cluster state change
+- register the Kyma Runtime within the CMP
```
klaudiagrz, line 11:
```suggestion
- Register Kyma Runtime within the CMP
```

concepts/infrastructure-reconciliation/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+# Proposed changes
+
+## Cluster Inventory as a separate service
+
+Cluster Inventory is the internal component responsible for reconciliation persistence. It keeps cluster information, configuration and reconciliation status. Component reconcilers do not have access to the inventory and cannot modify it directly. One option would be to extend action context passed to the reconciler to include more data, but the better option would be to pass only information how to access cluster inventory entry (url/id) and allow components reconciler to query and update it.
```
klaudiagrz, line 17:
```suggestion
Cluster Inventory is the internal component responsible for reconciliation persistence. It keeps cluster information, configuration, and reconciliation status. Component Reconcilers do not have access to the inventory and cannot modify it directly. One option would be to extend the action context passed to the Reconciler to include more data, but the better option would be to pass only information on how to access Cluster Inventory entry (URL/ID) and allow Components Reconciler to query and update it.
```

concepts/infrastructure-reconciliation/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Watches and triggers
+Reconciliation is currently time based (triggered in regular intervals). It is known limitation. In some cases reconciliation should be triggered immediately after some events, like:
```
klaudiagrz, line 20:
```suggestion
Reconciliation is currently time-based (triggered in regular intervals). It is a known limitation. In some cases, the reconciliation should be triggered immediately after some events, such as:
```

concepts/infrastructure-reconciliation/README.md
```diff
@@ -0,0 +1,39 @@
...
+- configuration change (e.g. network reconciler can add vnet name to the cluster inventory what should trigger shoot reconciler)
+- resource in the cluster modified (e.g. connectivity service instance created in the cluster should trigger connectivity proxy reconciler)
+
+Restoring managed resources accidentally (or intentionally) deleted by user is not a main goal of watchers. They should be used rather to monitor resources like service instances that should trigger some integration tasks in the Kyma Control Plane.
+
+Watchers and triggers can be more selective to limit the number of unnecessary invocations. Component reconcilers can specify which Cluster Inventory fields changes they want to be triggered with (i.e. Istio reconciler subscribing to the kubeconfig field).
```
klaudiagrz, line 26:
```suggestion
Watchers and triggers can be more selective to limit the number of unnecessary invocations. Component Reconcilers can specify which changes in the Cluster Inventory fields they would be triggered by (i.e. Istio Reconciler subscribing to the kubeconfig field). 
```

concepts/infrastructure-reconciliation/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+Watchers and triggers can be more selective to limit the number of unnecessary invocations. Component reconcilers can specify which Cluster Inventory fields changes they want to be triggered with (i.e. Istio reconciler subscribing to the kubeconfig field). 
+
+## No sequences or dependency management
+Reconciler tries to manage dependencies in the simple way. The prerequisites group installed first and then other components. Introducing infrastructure reconciler will make dependency management more complex. Moreover, dependencies are valid mainly for the first installation phase, and later on are not so important (in the second reconciliation eventing reconciler can be invoked before network reconciler). It should be responsibility of component reconciler to figure out if all dependencies exist and the reconciliation can be completed. If dependencies are not ready component reconciler should return the information that it is not ready and wait for the next trigger (time or event).
```
klaudiagrz, line 29:
```suggestion
Reconciler tries to manage dependencies in a simple way. The prerequisites group is installed first, before other components. Introducing Infrastructure Reconciler will make dependency management more complex. Moreover, dependencies are valid mainly for the first installation phase and are not so important later on (in the second reconciliation, Eventing Reconciler can be invoked before Network Reconciler). It should be the responsibility of the Component Reconciler to figure out if all dependencies exist and the reconciliation can be completed. If dependencies are not ready, Component Reconciler should return the information that it is not ready and wait for the next trigger (time or event).
```

concepts/infrastructure-reconciliation/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+# High Level Architecture
+
+![](assets/infrastructure-reconciliation.svg)
+
+Mothership reconciler should provide common functionality related to watching changes and triggering component reconcilers in response to events or with time based schedule. Such aspects like queueing reconciliation requests, consolidating them, updating reconciliation status, retries and back-off strategy should be implemented once (in the mothership).
```
klaudiagrz, line 35:
```suggestion
Mothership Reconciler should provide common functionality related to watching changes and triggering Component Reconcilers in response to events or with the time-based schedule. Such aspects as queueing reconciliation requests, consolidating them, updating reconciliation status, retries and back-off strategy should be implemented once (in the Mothership). 
```

concepts/infrastructure-reconciliation/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+# High Level Architecture
+
+![](assets/infrastructure-reconciliation.svg)
+
+Mothership reconciler should provide common functionality related to watching changes and triggering component reconcilers in response to events or with time based schedule. Such aspects like queueing reconciliation requests, consolidating them, updating reconciliation status, retries and back-off strategy should be implemented once (in the mothership). 
+
+Worker pool implementation has to be revisited. Mothership should not manage dependencies - component reconciler should be written in the way that can recognize missing dependencies and can fail fast. The next reconciliation invocation should happen when the configuration (cluster inventory) is updated (dependency). 
+
+Cluster Inventory should be extracted from the mothership and should have more flexible content. Component reconcilers should be able to read and update the content. Access control should be also established as we would need access to cluster inventory also from external components (e.g. Central Management Plane).
```
klaudiagrz, line 39:
```suggestion
Cluster Inventory should be extracted from the Mothership and should have more flexible content. Component Reconcilers should be able to read and update the content. Access control should be also established as we would need access to Cluster Inventory also from external components (e.g. Central Management Plane).
```

klaudiagrz (approved):
Since this is a concept, I'll turn a blind eye to the black-and-white diagram. However, consider adding it some color when you update it next time 🙂

PR #675, opened 2022-06-06
adjustments for configurable logging

a-thaler:
**Description**

- proposal for a revised configurable logging API
- ...
- ...

**Related issue(s)**

### DIFF
--- a/concepts/observability-configurable-logging/08-revised-api.md
+++ b/concepts/observability-configurable-logging/08-revised-api.md
@@ -34,7 +34,7 @@ We need a clear distinction between supported and tested setups, and unsupported
 ### Dedotting support
 
 There is no filter available for opensearch-typical dedotting. That is a mandatory feature and we planned to solve it with a Lua script. On the other hand, the Lua filter should be an unsupported feature. So having the convenience of dedotting support in the `LogPresetBinding` would result in a Lua script in the `LogPipeline`. Both ways will not be optimal.
--> Add dedotting support as a feature to the `LogPipeline`
+-> Add dedotting support as a feature to the `LogPipeline`.
 
 ### Meaningfulness of LogPreset
 
--- a/concepts/observability-configurable-logging/08-revised-api.md
+++ b/concepts/observability-configurable-logging/08-revised-api.md
@@ -1,0 +1,251 @@
+# Revised API
+
+## Motivation For Revise
+
+The basic idea outlined in [the concept for the custom operator](05-custom-fluentbit-operator.md) and [the concept for the servicecatalog integration](06-servicecatalog-integration.md) was to have
+1. a general layer supporting native fluent-bit config snippets. Out of that an overall fluent-bit config gets generated, validated and injected to the damonset
+2. a second layer on top, bringing the convinience of defining pipelines without knowing the underlaying fluent-bit concepts.
+
+The MVP of the overall feature was based on releasing a first version of the 1. native layer. While doing several problems were discovered while doing analysis of several aspects. The These problems adjusted the concepts slightley step by step without re-evaluating the overall picture once again. As a result more and more abstractions/oppinions were build into the first layer, making the layers not differentiated enough anymore. The flexibility on the first layer got lost while the second layer was not fitting anymore fully on top.
+
+## Considerations
+
+### Tag per pipeline mandatory
+
+Different pipelines will require different filters. By default there is only one tag like `kube.*` resulting from the central tail plugin. Every `LogPipeline` will require a dedicated `rewrite_tag` filter in order to apply some filtering, otherwise you will get indeterministic effects (the order of the filters is important).
+-> Every pipeline needs a rewrite_tag introducing a custom tag per pipeline. To generate filters later out of a `LogPreset`, the tag name needs to be known
+
+### Filesystem buffer mandatory
+
+We realized that using in-memory buffers will not decouple the pipelines well enough. If one output is down, the other output will stop working as well. It requires to enable a filesystem buffer for pipeline including size management to assure that the node filesystem gets not flooded
+-> Every pipeline needs a filesystem buffer, every output needs a storage size defaulting/validation
+
+### Match expressions should be limited
+
+Support of full match expressions in a `LogPipeline` will harm other pipelines as well as the central `tail` pipeline. People copying code snippets might have match expressions using `*`. It should be prevented.
+-> Match expressions should be defaulted/validated
+
+### Unsupported Plugins
+
+We need a clear distinction between supported/tested setups and unsupported setups. An HTTP plugin usage should be supported with any configuration while a stackdriver plugin usage is unsupported as we have no chance to cover testing.
+-> Introduce a mechanism to detect usage of unsupported plugins
+
+### Dedotting support
+
+There is no filter available for opensearch-typical dedotting. That is a mandatory feature and we planned to solve it via a lua script. On the other hand, the lua filter should be an unsupported feature. So having the convinience of detotting support in the `LogPresetBinding` would result in a lua script in the `LogPipeline`.
+-> Add dedotting support as a feature to the `LogPipeline`
+
+### Meaningfulness of LogPreset
+
+It turned out that an OpenSearch integration via a HTTP plugin required for an internal SAP scenario can be much simpler then expected. A real templating is not needed and probably a plain simple pipeline config will be better suited. Also, what presets to provide, we cannot think of other scenarios which can be delivered. The binding to a secret can be done already in the pipeline itself, workload selectors could be added as well
+-> do we need the LogPreset concept?
+
+### Support of additional inputs
+It turned out that there are typical inputs available which you would like to process as well, mainly the kubelet logs
+-> add a source selector
+
+## Revised API
+
+More and more oppinion/abstraction was added to the `LogPipeline` concept and the meaningfullnes of the `LogPreset` is in the room. We could now try to push down the `LogPipeline` to be more native and accepting that the user can do really stupid things in combination of supported plugins and revise the LogPreset concept.
+Or we push up even more the `LogPipeline` API by introducing the convinience of the LogPreset while keeping native support as a customization option clearly indicating an unsupported scenario.
+This proposal favors the second idea by dropping the LogPreset fully and having only one layer which is already an abstraction. Giving no chance to configure fluent-bit natively. For that you can simply go with a custom installation.
+
+```YAML
+kind: LogPipeline
+apiVersion: telemetry.kyma-project.io/v1alpha1
+metadata:
+  name: OpenSearchHTTP-App
+spec:
+  input:
+    containers: # maps to central tail pipeline
+      namespaces:[] # maps to rewrite_tag rule
+      excludeNamespaces: []
+      pods: []
+      containers: []
+    system: {} #maps to systemd based input like kubelet logs
+    custom: | # entering unsupported mode
+      Name dummy
+      Dummy {"message":"dummy"}
+  filters:
+    - parser: "json" # support of built-in parsers
+    - dedot: {}
+    - modify:
+        operation: "add"
+        key: cluster_identifier
+        value: ${KUBERNETES_SERVICE_HOST}
+    - modify:
+        operation: "remove"
+        key: cluster_identifier
+    - select:
+        include:
+          key:
+          regexp:
+        exclude:
+          key:
+          regexp:
+    - custom: | # no "Match" available, entering unsupported mode
+        Name    record_modifier
+        Record  cluster_identifier ${KUBERNETES_SERVICE_HOST}
+  output: #only one output
+    http:
+      Host: ${ES_ENDPOINT}
+      HTTP_User: ${ES_USER} # Defined in Secret
+      HTTP_Password: ${ES_PASSWORD} # Defined in Secret
+    custom: | # no "Match" available, no filebuffer settings available, entering unsupported mode
+        Name               es
+        Alias              es-output
+        Host               ${ES_ENDPOINT} # Defined in Secret
+        HTTP_User          ${ES_USER} # Defined in Secret
+        HTTP_Password      ${ES_PASSWORD} # Defined in Secret
+        LabelMapPath       /files/labelmap.json
+   variables:
+      - fromSecretRef:
+            name: my-elastic-credentials
+            namespace: default
+      - fromSecretPrefixRef: # secret rotation
+            name: my-elastic-credentials
+            namespace: default
+      - fromSecretKeyRef:
+            name: my-elastic-credentials
+            namespace: default
+            key: ES_ENDPOINT
+      - fromConfigMapRef:
+            name: my-elastic-config
+            namespace: default
+      - fromConfigMapKeyRef:
+            name: my-elastic-credentials
+            namespace: default
+            key: ES_ENDPOINT
+
+  files:
+    - name: labelmap.json
+      content: |
+      {
+          "kubernetes": {
+            "namespace_name": "namespace",
+            "pod_name": "pod"
+          },
+          "stream": "stream"
+      }
+   - fromConfigMapRef:
+       name: my-loki-labelMap
+       namespace: default
+   - fromSecretRef:
+       name: my-ip-whitelist
+       namespace: default
+  parsers:
+   - custom: |
+        Name dummy_test
+        Format regex
+        Regex ^(?<INT>[^ ]+) (?<FLOAT>[^ ]+) (?<BOOL>[^ ]+) (?<STRING>.+)$
+  multilineParsers:
+    - custom: |
+        # Example from https://docs.fluentbit.io/manual/pipeline/filters/multiline-stacktrace
+        name          multiline-custom-regex
+        type          regex
+        flush_timeout 1000
+        rule      "start_state"   "/(Dec \d+ \d+\:\d+\:\d+)(.*)/"  "cont"
+        rule      "cont"          "/^\s+at.*/"                     "cont" 
+
+```
+
+Example of typical OpenSearch HTTP Application Log pipeline:
+```YAML
+kind: LogPipeline
+apiVersion: telemetry.kyma-project.io/v1alpha1
+metadata:
+  name: OpenSearchHTTP-App
+spec:
+   input:
+      containers:
+         excludeNamespaces: ["kyma-system", "kube-system"]
+   filters:
+      - parser: "json"
+      - dedot: {}
+      - modify:
+         operation: "add"
+         key: cluster_identifier
+         value: ${KUBERNETES_SERVICE_HOST}
+   output:
+      http:
+         Host: ${ES_ENDPOINT}
+         Port: 443
+         HTTP_User: ${ES_USER} # Defined in Secret
+         HTTP_Password: ${ES_PASSWORD} # Defined in Secret
+         URI: /customindex/kyma
+         Format: json
+         allow_duplicated_headers: true
+
+   variables:
+      - fromSecretKeyRef:
+            namespace: default
+            prefix: my-elastic
+            key: HTTP_USER
+            variable: ES_USER
+      - fromSecretKeyRef:
+            namespace: default
+            prefix: my-elastic
+            key: HTTP_PASSWORD
+            variable: ES_PASSWORD
+      - fromSecretKeyRef:
+            namespace: default
+            prefix: my-elastic
+            key: HTTP_HOST
+            variable: ES_ENDPOINT
+```
+
+Example of typical OpenSearch HTTP Istio Access Log pipeline:
+```YAML
+kind: LogPipeline
+apiVersion: telemetry.kyma-project.io/v1alpha1
+metadata:
+  name: OpenSearchHTTP-App
+spec:
+   input:
+      containers:
+         excludeNamespaces: ["kyma-system", "kube-system"]
+         containers: ["istio-proxy"]
+   filters:
+      - parser: "json"
+      - dedot: {}
+      - modify:
+            operation: "add"
+            key: cluster_identifier
+            value: ${KUBERNETES_SERVICE_HOST}
+      - select:
+         include:
+            key: protocol
+            regexp: ".+"
+   output:
+      http:
+         Host: ${ES_ENDPOINT}
+         Port: 443
+         HTTP_User: ${ES_USER} # Defined in Secret
+         HTTP_Password: ${ES_PASSWORD} # Defined in Secret
+         URI: /customindex/istio-envoy-kyma
+         Format: json
+         allow_duplicated_headers: true
+
+   variables:
+      - fromSecretKeyRef:
+            namespace: default
+            prefix: my-elastic
+            key: HTTP_USER
+            variable: ES_USER
+      - fromSecretKeyRef:
+            namespace: default
+            prefix: my-elastic
+            key: HTTP_PASSWORD
+            variable: ES_PASSWORD
+      - fromSecretKeyRef:
+            namespace: default
+            prefix: my-elastic
+            key: HTTP_HOST
+            variable: ES_ENDPOINT
+```
+
+## Pros
+
+- Approach makes one common layer which is clearly oppinionated but has options to opt-out in a controlled way (match expressions will still be enforced)
+
+## Cons
--- a/concepts/observability-configurable-logging/08-revised-api.md
+++ b/concepts/observability-configurable-logging/08-revised-api.md
@@ -46,70 +46,78 @@ It turned out that there are typical inputs available which you would like to pr
 
 ## Revised API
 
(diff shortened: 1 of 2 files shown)


concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,290 @@
...
+-> Every pipeline needs a filesystem buffer, every output needs a storage size defaulting/validation
+
+### Match expressions should be limited
+
+Support of full match expressions in a `LogPipeline` will harm other pipelines as well as the central `tail` pipeline. People copying code snippets might have match expressions using `*`. It should be prevented.
+-> Match expressions should be defaulted/validated
+
+### Unsupported Plugins
+
+We need a clear distinction between supported/tested setups and unsupported setups. An HTTP plugin usage should be supported with any configuration while a stackdriver plugin usage is unsupported as we have no chance to cover testing.
```
chrkl, line 30: The additional resource consumption of some plugins and settings is also a relevant factor. We don't want the user to configure anything that creates a new emitter, which increases the memory usage above the amount that we introduce ourselves.

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,290 @@
...
+      containers: []
+      excludeContainers: []
+      matchLabels:
+        app: icke
+    system: {} #maps to systemd based input like kubelet logs
+    custom: | # entering unsupported mode
+      Name dummy
+      Dummy {"message":"dummy"}
+  filters:
+    - multilineParser:
```
chrkl, line 74: The alternative of defining parsers and multiline parsers in the filters section would be a global configuration, for instance through additional `Parser` and `MultilineParser` CRDs. Parsers could be used by annonations and multiline parsers applied directly on the input.

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,290 @@
...
+    custom:
+      | # no "Match" available, no filebuffer settings available, entering unsupported mode
+      Name               es
+      Alias              es-output
+      Host               ${ES_ENDPOINT} # Defined in Secret
+      HTTP_User          ${ES_USER} # Defined in Secret
+      HTTP_Password      ${ES_PASSWORD} # Defined in Secret
+      LabelMapPath       /files/labelmap.json
+
+  custom:
```
chrkl, line 121: Not sure if custom is a good name here. It might give the impression that everything here is unsupported. Should custom parsers be unsupported?

concepts/observability-configurable-logging/08-revised-api.md
a-thaler: So maybe you can define multilineParsers inline on the input? Of course then it is not re-usable but maybe that is acceptable?

```
input:
  application:
    multilineParsers:
      - type: "java"
      - custom: |
        XXX
```

Same we could do for parsers, inline on the filer. However, then you cannot re-use it on the annotation..

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,290 @@
...
+    application: # maps to central tail pipeline
+      namespaces: [] # maps to rewrite_tag rule
+      excludeNamespaces: []
+      pods: []
+      excludePods: []
+      containers: []
+      excludeContainers: []
+      matchLabels:
+        app: icke
+    system: {} #maps to systemd based input like kubelet logs
```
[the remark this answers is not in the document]
a-thaler: I have not dealt with that logs at all yet, can you make a proposal? Anyway we will not implement it as a first thing, still it will be good to understand the potential API already

concepts/observability-configurable-logging/08-revised-api.md
a-thaler: If there is something supported, then it should not below that, I agree. About parsers... can a custom parser be harmful? Probably not.. but also goes back to the comment above about moving parsers into an own CRD. Introducing a Parsers CRD sounds very fluentbit specific and should cover other central configs in future.. maybe we can keep them for now in the pipeline itself and compile the central config out of it till we have a better understanding on what else will be central?

So maybe let's keep all as non-custom indeed. Defining parsers/multilineparsers will not be harmful at all as long as you don't use them, same for variables and files.

concepts/observability-configurable-logging/08-revised-api.md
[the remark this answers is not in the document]
chrkl: The documentation of the systemd input gives the example to use the Systemd_Filter property to select a specific unit. The use-case that I have seen before is to limit the input the the kubelet unit. Giving a list of units is the only use-case that I see at the moment. But I'm fine to postpone the decision on details.

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,290 @@
...
+        exclude:
+          key:
+          regexp:
+    - custom: | # no "Match" available, entering unsupported mode
+        Name    record_modifier
+        Record  cluster_identifier ${KUBERNETES_SERVICE_HOST}
+
+  output: #only one output
+    http:
+      Dedot: true
```
rakesh-garimella, line 98:
```suggestion
      Dedot: true
      Match: Logpipeline.*
```

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,290 @@
...
+          key:
+          regexp:
+        exclude:
+          key:
+          regexp:
+    - custom: | # no "Match" available, entering unsupported mode
+        Name    record_modifier
+        Record  cluster_identifier ${KUBERNETES_SERVICE_HOST}
+
+  output: #only one output
```
rakesh-garimella, line 96: Just a though can we can may be have individual crds for each output we would support. With this we can introduce custom outputs which would be wrapper around basic outputs (like http). I find `dedot` under http a bit misleading.

a-thaler: I don't get the idea with the wrapping. So you suggest a new CRD `Output` and in the pipeline you just reference a concrete resource of that type which is already filled with concrete parameters or still generic?

Can you just outline quickly an example to get the idea more easy

concepts/observability-configurable-logging/08-revised-api.md
a-thaler: The idea was to unburden the user from understand tagging in fluent-bit at all. So you are limited to the logs of the pipeline selected in the inout section, that's it. What is your motivation of having match configurable? To support applying it to a subset of logs only? We dropped that idea, elements  will be applied to all in the pipeline always. We also suggest to have only one output in a pipeline (it is singular now), so applying an output to a subset makes no sense anyway then as you can just limit the logs already in the input section.

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,279 @@
...
+          name: my-elastic-credentials
+          namespace: default
+          key: ES_ENDPOINT
+      HTTP_Password:
+        fromSecretKeyRef:
+          name: my-elastic-credentials
+          namespace: default
+          key: ES_ENDPOINT
+    custom:
+      | # no "Match" available, no filebuffer settings available, entering unsupported mode
```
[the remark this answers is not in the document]
a-thaler: The idea is that the match expression cannot be adjusted at all, see also reasoning in your comment above.

Buffer sizes we can think of allowing custom settings as long as we keep the limits in place. Or we just admit that this is unsupported mode and user can do what he wants. Maybe we can start with 2. and later implement 1. option. What do you think?

concepts/observability-configurable-logging/08-revised-api.md
rakesh-garimella:
```
kind: http-special
apiVersion: outpuputs.telemetry.kyma-project.io/v1alpha1
metadata:
  name: myhttp
spec:
     dedot: true
      Host:
        value: "icke.com"
      HTTP_User:
        value: "icke"
        fromSecretKeyRef:
          name: my-elastic-credentials
          namespace: default
          key: ES_ENDPOINT
      HTTP_Password:
        fromSecretKeyRef:
          name: my-elastic-credentials
          namespace: default
          key: ES_ENDPOINT
---
  kind: http
apiVersion: outpuputs.telemetry.kyma-project.io/v1alpha1
metadata:
  name: myhttp
spec:
      Host:
        value: "icke.com"
      HTTP_User:
        value: "icke"
        fromSecretKeyRef:
          name: my-elastic-credentials
          namespace: default
          key: ES_ENDPOINT
      HTTP_Password:
        fromSecretKeyRef:
          name: my-elastic-credentials
          namespace: default
          key: ES_ENDPOINT
```

concepts/observability-configurable-logging/08-revised-api.md
a-thaler: We agreed to extract parsers and multiLineParsers into a dedicated CRD

concepts/observability-configurable-logging/08-revised-api.md
a-thaler: custom on root level removed

concepts/observability-configurable-logging/08-revised-api.md
a-thaler: We agreed to not extract the output into a dedicated CRD as we don't want to model the SAP specific scenario of a OpenSearchHTTP output explicitly but stay with common http plugin. Also, a re-use scenario we cannot foresee for now.

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,298 @@
...
+## Considerations
+
+### Tag per pipeline mandatory
+
+Different pipelines will require different filters. By default there is only one tag like `kube.*` resulting from the central tail plugin. Every `LogPipeline` will require a dedicated `rewrite_tag` filter in order to apply some filtering, otherwise you will get indeterministic effects (the order of the filters is important).
+-> Every pipeline needs a rewrite_tag introducing a custom tag per pipeline. To generate filters later out of a `LogPreset`, the tag name needs to be known
+
+### Filesystem buffer mandatory
+
+We realized that using in-memory buffers will not decouple the pipelines well enough. If one output is down, the other output will stop working as well. It requires to enable a filesystem buffer for pipeline including size management to assure that the node filesystem gets not flooded
```
NHingerl, line 20:
```suggestion
We realized that using in-memory buffers will not decouple the pipelines well enough. If one output is down, the other output will stop working as well. To assure that the node filesystem is not flooded, there must be an enabled filesystem buffer for each pipeline, including size management.
```
Just guessing the "each".
We could probably drop the word "enabled".

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,298 @@
...
+-> Every pipeline needs a filesystem buffer, every output needs a storage size defaulting/validation
+
+### Match expressions should be limited
+
+Support of full match expressions in a `LogPipeline` will harm other pipelines as well as the central `tail` pipeline. People copying code snippets might have match expressions using `*`. It should be prevented.
+-> Match expressions should be defaulted/validated
+
+### Unsupported Plugins
```
NHingerl, line 28:
```suggestion
### Unsupported plugins
```

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,298 @@
...
+
+We need a clear distinction between supported/tested setups and unsupported setups. An HTTP plugin usage should be supported with any configuration while a stackdriver plugin usage is unsupported as we have no chance to cover testing. The additional resource consumption of some plugins and settings is also a relevant factor. We don't want the user to configure anything that creates a new emitter, which increases the memory usage above the amount that we introduce ourselves.
```
NHingerl, line 30:
```suggestion
We need a clear distinction between supported and tested setups, and unsupported setups. Using an HTTP plugin should be supported with any configuration, while a stackdriver plugin is unsupported because we have no chance to cover testing. The additional resource consumption of some plugins and settings is also a relevant factor. We don't want the user to configure anything that creates a new emitter, which increases the memory usage above the amount that we introduce ourselves.
```

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,298 @@
...
+-> Introduce a mechanism to detect usage of unsupported plugins
```
NHingerl, line 31:
```suggestion
-> Introduce a mechanism to detect usage of unsupported plugins.
```

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,298 @@
...
+
+### Dedotting support
+
+There is no filter available for opensearch-typical dedotting. That is a mandatory feature and we planned to solve it via a lua script. On the other hand, the lua filter should be an unsupported feature. So having the convinience of detotting support in the `LogPresetBinding` would result in a lua script in the `LogPipeline`.
```
NHingerl, line 35:
```suggestion
There is no filter available for opensearch-typical dedotting. That is a mandatory feature and we planned to solve it with a Lua script. On the other hand, the Lua filter should be an unsupported feature. So having the convenience of dedotting support in the `LogPresetBinding` would result in a Lua script in the `LogPipeline`.
```

So on the one hand, we forbid Lua scripts/filter, but we planned to use one ourselves; which is not ideal, right? I'm not sure if the conclusion in line 36 now means that we're not going to use a Lua script and instead find some other solution for dedotting support, or that we're going to add the Lua script as a feature to the LogPipeline. Can we add a short sentence to make it clearer which solution we chose?

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,298 @@
...
+-> Add dedotting support as a feature to the `LogPipeline`
```
NHingerl, line 36:
```suggestion
-> Add dedotting support as a feature to the `LogPipeline`.
```

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,298 @@
...
+
+### Meaningfulness of LogPreset
+
+It turned out that an OpenSearch integration via a HTTP plugin required for an internal SAP scenario can be much simpler then expected. A real templating is not needed and probably a plain simple pipeline config will be better suited. Also, what presets to provide, we cannot think of other scenarios which can be delivered. The binding to a secret can be done already in the pipeline itself, workload selectors could be added as well
```
NHingerl, line 40: I don't understand this sentence: "Also, what presets to provide, we cannot think of other scenarios which can be delivered." Just guessing here:

```suggestion
It turned out that an OpenSearch integration with an HTTP plugin required for an internal SAP scenario can be much simpler than expected. A real templating is not needed, and probably a plain simple pipeline config will be better suited. Also, regarding presets to provide, we cannot think of other scenarios that can be delivered. The binding to a secret can be done already in the pipeline itself, workload selectors could be added as well.
```

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,298 @@
...
+-> do we need the LogPreset concept?
```
NHingerl, line 41:
```suggestion
-> Do we need the LogPreset concept?
```

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,298 @@
...
+
+### Support of additional inputs
+It turned out that there are typical inputs available which you would like to process as well, mainly the kubelet logs
```
NHingerl, line 44:
```suggestion
It turned out that there are typical inputs available (mainly the kubelet logs) that users might want to process as well.
```

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,298 @@
...
+-> add a source selector
```
NHingerl, line 45:
```suggestion
-> Add a source selector.
```

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,298 @@
...
+
+## Revised API
+
+More and more oppinion/abstraction was added to the `LogPipeline` concept and the meaningfullnes of the `LogPreset` is questioned. We could try to push down the `LogPipeline` on the scala of abstraction to be more native and accepting that the user can do really stupid things in combination of supported plugins and revise the `LogPreset` concept.
```
NHingerl, line 49:
```suggestion
More and more opinion and abstraction was added to the `LogPipeline` concept, and the meaningfulness of the `LogPreset` is questioned. We came up with the following ideas:
- We could try to push down the `LogPipeline` on the scala of abstraction to be more native (accepting that the user might do really stupid things by combining supported plugins), and revise the `LogPreset` concept.
```

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,298 @@
...
+Or we even more push up the `LogPipeline` on the scala by introducing the convinience of the `LogPreset` while keeping native support as a customization option clearly indicating an unsupported scenario.
```
NHingerl, line 50:
```suggestion
- Alternatively, we push up the `LogPipeline` on the scala of abstraction even more by introducing the convenience of the `LogPreset`, while keeping native support as a customization option, clearly indicating an unsupported scenario.
```

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,298 @@
...
+
+This proposal favors the second idea by dropping the `LogPreset` fully and keeping one layer only which is focused on an abstraction and covering the actual value for users. A pure native layer per se brings no real value to users, as the user needs to understand the fluentbit concepts and the specifics of the layer gaining only the lifecycle management aspects of the fluentbit instance (but what kind of guarantees in regards to support).
```
NHingerl, line 52:
```suggestion
This proposal favors the second idea by dropping the `LogPreset` fully and keeping only one layer that is focused on an abstraction and covers the actual value for users. A pure native layer per se adds no real value for users, as the user must understand the Fluent Bit concepts and the specifics of the layer, gaining only the lifecycle management aspects of the Fluent Bit instance (but what kind of guarantees in regards to support). 
```

What does this mean? "but what kind of guarantees in regards to support"

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,298 @@
...
+Or we even more push up the `LogPipeline` on the scala by introducing the convinience of the `LogPreset` while keeping native support as a customization option clearly indicating an unsupported scenario.
+
+This proposal favors the second idea by dropping the `LogPreset` fully and keeping one layer only which is focused on an abstraction and covering the actual value for users. A pure native layer per se brings no real value to users, as the user needs to understand the fluentbit concepts and the specifics of the layer gaining only the lifecycle management aspects of the fluentbit instance (but what kind of guarantees in regards to support). 
+
+The proposal takes the existing `LogPipeline` and extends it with a more abstracted syntax, still keeping the option to customize single elements. As parsers are a central config element, they get extracted into a dedicated resource `LogParser`
```
NHingerl, line 54:
```suggestion
The proposal takes the existing `LogPipeline` and extends it with a more abstracted syntax, still keeping the option to customize single elements. As parsers are a central config element, they are extracted into a dedicated resource `LogParser`.
```

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,298 @@
...
+          key: "label.json"
+    - name: "cert.crt"
+      contentFrom:
+        secretKeyRef:
+          name: my-certificates
+          namespace: default
+          key: "cert.crt"
+```
+
+A `LogParser` specifies exactly one parser or a multilineparser and gets activated instantly on the tail plugin (multilineParser) or to be used in an annotation (parser)
```
NHingerl, line 174:
```suggestion
A `LogParser` specifies exactly one parser or a multiline parser, and is activated instantly on the tail plugin (multiline parser) or to be used in an annotation (parser).
```

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,298 @@
...
+
+```YAML
+kind: LogParser # cluster scope
+apiVersion: telemetry.kyma-project.io/v1alpha1
+metadata:
+  name: multiline-custom-regex
+spec:
+  parser: # Name is rejected as it gets generated out of resource name
```
NHingerl, line 182:
```suggestion
  parser: # Name is rejected because it is generated from the resource name
```

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,298 @@
...
+          # Will be registered as parser to be used in annotations or in a pipeline via a custom parser filter
```
NHingerl, line 183:
```suggestion
          # Will be registered as parser to be used in annotations, or in a pipeline using a custom parser filter
```

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,298 @@
...
+      |
+        Format regex
+        Regex ^(?<INT>[^ ]+) (?<FLOAT>[^ ]+) (?<BOOL>[^ ]+) (?<STRING>.+)$
+  multilineParser: # Name is rejected as it gets generated out of resource name
```
NHingerl, line 187:
```suggestion
  multilineParser: # Name is rejected because it is generated from the resource name
```

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,298 @@
...
+```
+
+Example of minimal OpenSearch HTTP Application Log pipeline:
+```YAML
+kind: LogPipeline
+apiVersion: telemetry.kyma-project.io/v1alpha1
+metadata:
+  name: OpenSearchHTTP-App
+spec:
+  input: #  optional section as application is default and all logs are processed by default
```
NHingerl, line 203:
```suggestion
  input: #  optional section because application is default and all logs are processed by default
```

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,298 @@
...
+```
+
+Example of typical OpenSearch HTTP Istio Access Log pipeline:
+```YAML
+kind: LogPipeline
+apiVersion: telemetry.kyma-project.io/v1alpha1
+metadata:
+  name: OpenSearchHTTP-Istio
+spec:
+  input: # mandatory section to include only istio containers
```
NHingerl, line 241:
```suggestion
  input: # mandatory section to include only Istio containers
```

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,298 @@
...
+    application:
+      excludeNamespaces: ["kyma-system", "kube-system"]
+      containers: ["istio-proxy"]
+  filters: # mandatory section to exclude logs having no protocol
```
NHingerl, line 245:
```suggestion
  filters: # mandatory section to exclude logs that have no protocol
```

concepts/observability-configurable-logging/08-revised-api.md
a-thaler: Semantic was wrong, re-phrased your suggestion to have it easy to merge

concepts/observability-configurable-logging/08-revised-api.md
a-thaler: Thats exactly the dilemma, there is no good solution with the old approach

NHingerl: Aaaah!

concepts/observability-configurable-logging/08-revised-api.md
```diff
@@ -0,0 +1,300 @@
...
+-> Add a source selector.
+
+## Revised API
+
+More and more opinion and abstraction was added to the `LogPipeline` concept, and the meaningfulness of the `LogPreset` is questioned. We came up with the following ideas:
+- We could try to push down the `LogPipeline` on the scala of abstraction to be more native (accepting that the user might do really stupid things by combining supported plugins), and revise the `LogPreset` concept.
+- Alternatively, we push up the `LogPipeline` on the scala of abstraction even more by introducing the convenience of the `LogPreset`, while keeping native support as a customization option, clearly indicating an unsupported scenario.
+
+This proposal favors the second idea by dropping the `LogPreset` fully and keeping only one layer that is focused on an abstraction and covers the actual value for users. A pure native layer per se adds no real value for users, as the user must understand the Fluent Bit concepts and the specifics of the layer, gaining only the lifecycle management aspects of the Fluent Bit instance (but what kind of guarantees he will get from the lifecycle management in regards to support?).
```
NHingerl, line 54:
```suggestion
This proposal favors the second idea by dropping the `LogPreset` fully and keeping only one layer that is focused on an abstraction and covers the actual value for users. A pure native layer per se adds no real value for users, as the user must understand the Fluent Bit concepts and the specifics of the layer, gaining only the lifecycle management aspects of the Fluent Bit instance - however, the lifecycle management does not provide any clearly guaranteed support.
```

NHingerl (approved):

PR #705, opened 2022-08-26
Respect CLS requirements in the log format recommendations

kwiatekus:
**Description**

- timestamp->date
- message->msg
- avoid nesting in the context object

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/14873


concepts/observability-consistent-logging/improvement-of-log-messages-usability.md
```diff
@@ -28,15 +28,15 @@ Additional assumptions:
...
+- **context** - structure of the contextual information, such as operation (for example: `starting workers`), handler/ resolver (for example: `ProvisionRuntime`), controller, resource-namespaced name (for example: `production/application1`), operation ID, instance ID, operation stage, and so on. Users must be able to filter the logs so all the info provided here must be a useful and unique minimal set for every operation. Users must be able to find the needed resource in some store so provide here a name instead of an ID if it's easier to use later. Flat context object is recommended as in the end the the cloud storage flattens it to a key/value map to make it searchable via OpenSearch. Avoid dots (`.`) in the keys.
```
grego952, line 39:
```suggestion
- **context** - structure of the contextual information, such as operation (for example: `starting workers`), handler/ resolver (for example: `ProvisionRuntime`), controller, resource-namespaced name (for example: `production/application1`), operation ID, instance ID, operation stage, and so on. Users must be able to filter the logs so all the info provided here must be a useful and unique minimal set for every operation. Users must be able to find the needed resource in some store to provide here a name instead of an ID if it's easier to use later. Flat context object is recommended as in the end the cloud storage flattens it to a key-value map to make it searchable using OpenSearch. Avoid dots (`.`) in the keys.
```

grego952 (approved):

PR #714, opened 2022-09-12
Cluster Operator proposal

janmedrek:
**Description**

- Introduce the Cluster Operator, Subscription Operator and Kubeconfig Operator components.
- Provide minimal, exemplary `Cluster`, `SubscriptionRequest`, and `KubeconfigRequest` CRs.
- Create the high-level architecture diagram of the infrastructure reconciliation area.

**Related issue(s)**
https://github.com/kyma-project/lifecycle-manager/issues/12
https://github.com/kyma-project/lifecycle-manager/issues/119


concepts/infrastructure-reconciliation/README.md
```diff
@@ -1,41 +1,172 @@
...
+- SubscriptionRequest CR
+- KubeconfigRequest CR
+
+### Cluster Operator and the Cluster CR
+
+The Cluster Operator's main responsibility is to manage the Cluster CRs, request subscriptions for the provisioning and create Shoot CRs in the Gardener cluster.
+
+On top of that, there is also a requirement to provide out-of-the-box admin roles to some users, this is the only place where the Cluster Operator interacts directly with the created cluster.
+
+Also, the exemplary Cluster CR resource looks as follows:
```
jakobmoellersap, line 64: QQ: Is there a document somewhere explaining why we do not use Kubernetes Cluster SIG? I would have imagined you could simply write a cluster class: https://cluster-api.sigs.k8s.io/tasks/experimental-features/cluster-class/write-clusterclass.html

jakobmoellersap (commented):
In General, I am missing a Reconciliation Flow for creation of a Kyma CR within the Cluster Operator. I think in general the only way to couple this right now is by having an additional control loop in cluster operator or lifecycle manager that creates Kyma CRs based on the Infrastructure provided in the Cluster CR. It would be great if this workflow could be roughly outlined.

Also minor nit. If there is lifecycle-manager, module-manager, probably it could be `cluster-manager` instead of operator

pbochynski (commented):
Can you add example with Bring Your Own Cluster model? How KubeconfigRequest would look like in that case? Do we have Cluster resource for BYOC model that has just a kubeconfig inside, or something else?

concepts/infrastructure-reconciliation/README.md
janmedrek: I've read a bit into that and I find the resources described in the Kubernetes Cluster SIG more infrastructure-focused.

Here, we want to somehow unite both the infrastructure and business aspects of the cluster provisioning. We already have a Gardener's Shoot CRs for the low-level infrastructure configuration, I would rather like to abstract from that and provide a simple Cluster API, while hiding the network/worker config underneath - we can contain that inside the Cluster Manager's logic.

Anyways, after few meetings the idea is to use some of the patterns that Kubernets Cluster SIG utilizes when it comes to the K8S resources and how they are tied to eachother.

janmedrek:
@jakobmoellersap I've had a quick follow-up with @pbochynski on that design, this is a rough sketch of what we came up with. I've also included the "full-flow" only for the managed cluster to keep the diagram a bit simpler, based on [this document](https://github.com/kyma-project/community/tree/main/concepts/modularization#example-for-local-and-central-operators):

infrastructure-full

The idea is to extend the Cluster CR with the references to two different types of resources - CustomerCluster and GardenerCluster (names TBD), those require different types of params and there is no pretty solution that uses just one CR for both types of clusters. This way there will be both a separation between APIs for different licence types and a way to list all clusters regardless of their type.

As a bonus, we can get rid of Kubeconfig Manager and have separate operators for different flows, simplifying the overall architecture of the solution.

Let me know what you think - I know there is a meeting scheduled to discuss this idea next week, perhaps you may have any comments a bit earlier. I will be updating the proposal as soon as we have a more defined direction established.

concepts/infrastructure-reconciliation/README.md
```diff
@@ -1,41 +1,196 @@
...
+- SubscriptionRequest CR
+
+### Cluster Managers and the Cluster CRs
+
+The Cluster Managers' main responsibility is to manage the respective Cluster CRs, this is tied to both their provisioning and providing an access to the created K8S clusters.
+
+First of all, an overarching Cluster CR will provide a common API for operations connected to support or infrastructure maintenance. It contains reference to the resource corresponding to the sub-type of cluster:
+```
+apiVersion: operator.kyma-project.io/v1alpha1
+kind: Cluster
```
jakobmoellersap, line 74: I think this kind of cluster API is bound to not cover all cases and we should not introduce a common Cluster CR if we anyhow have two separate CRs (GardenerCluster and CustomerCluster). To be frank, all this does is hold a ref to another CR, so why even reconcile it instead of reconciling Customer/Gardener Clusters directly?

concepts/infrastructure-reconciliation/README.md
```diff
@@ -1,41 +1,196 @@
...
+```
+
+This contains a minimal set of values that are required to provision a new Gardener cluster. The resource also contains the `administrators` and `subscriptionSelector` fields that are required for the Binding creation and requesting a hyperscaler subscription respectively.
+
+And the second one for BYOC model:
+```
+customer-cluster
+
+apiVersion: operator.kyma-project.io/v1alpha1
+kind: CustomerCluster
```
jakobmoellersap, line 138:
```suggestion
kind: ExternalCluster
```

jakobmoellersap: I think CustomerCluster is not really fitting the role here, but its just a nit, so nothing urgent

concepts/infrastructure-reconciliation/README.md
```diff
@@ -1,41 +1,196 @@
...
+		- apiGroups: ["*"]
+		  resources: ["*"]
+		  verbs: ["*"]
+```
+
+This is the simpler one, as there is no cluster provisioning happening in this flow most of the parameters can be dropped.
+
+### Issuing kubeconfigs
+
+As stated before, the separation of kubeconfigs and limiting access is needed to have a healthy, secure environment. Cluster Managers will provide a possibility to create RBAC-based kubeconfigs in both BYOC and Gardener clusters. This will provide operators with the required access to perform any necessary actions.
```
jakobmoellersap, line 157: I think a major flaw that the entire environment has is that operators managing other operator deliveries need almost administrative access since to avoid privilege escalations, they must hold all permissions they are trying to grant to an operator they deploy. This means that even though we can create RBAC-based kubeconfigs, module manager most likely will request (almost) full administrative access. Maybe something to discuss separately, but worth mentioning. @adityabhatia FYI this is what we observed in our POCs around integrating module manager in real life operators with tons of crds and RBACs

concepts/infrastructure-reconciliation/README.md
```diff
@@ -1,41 +1,196 @@
...
 
-Worker pool implementation has to be revisited. Mothership should not manage dependencies - Component Reconciler should be written in a way that can recognize missing dependencies and fail fast. The next reconciliation invocation should happen when the configuration (Cluster Inventory) is updated (dependency). 
+After the CRs are created, the respective Cluster Manager will pick up the provisioning request and will create a new Shoot for the runtime (managed scenario only), and provide a requested kubeconfig.
 
-Cluster Inventory should be extracted from the Mothership and should have more flexible content. Component Reconcilers should be able to read and update the content. Access control should be also established as we would need access to Cluster Inventory also from external components (e.g. Central Management Plane).
\ No newline at end of file
+At this point Lifecycle Manager can step in, it has all required resources to set up the Runtime:
+- running cluster
+- kubeconfig (passed via secret reference)
```
jakobmoellersap, line 195: How do we listen for secret rotation? Via Secret Resource Version change or via A change on the status of the Cluster APIs. I think looking at 5k+ Clusters, Resource Version Changes for 5k+ Secrets are gonna be heavy on the watches for the API server and provide a bottleneck. Is there gonna be a generational interface on the clusters we can use to reduce the change amount we need to filter on downstream in lifecycle manager?

concepts/infrastructure-reconciliation/README.md
jakobmoellersap: Also, if we want to keep Cluster as general API, I think this is an incorrect implementation since you bind the Interface API (Kind Cluster) with different Ref types to their implementations (gardenerClusterRef and customerClusterRef). Looking at Kubernetes upstream, this mistake was done multiple times in API Design (e.g. in the design phase of PVs and StorageClass) and they had to redo this.

Looking at the learnings, I think we should follow a similar approach to the upcoming Kubernetes Gateway API (see https://gateway-api.sigs.k8s.io/api-types/gatewayclass/):

This would mean that we have 1 ClusterClass for each of the implementations and an Interface API that can deal with interpreting calls from a ClusterClass. Then the Cluster Controller Implementations only have to be able to deal with the ClusterClass implementation by using a field called "clusterClassName: Gardener" in the Spec of the Cluster. That way you do not have to introduce new Ref Types for a new CRD you want to add as implementation detail.

Let me know what you think!

jakobmoellersap (changes requested):

concepts/infrastructure-reconciliation/README.md
janmedrek: The goal was to not reconcile the common Cluster CR at all - rather keep it as a way to simplify the operations regarding environment management, support or statistics. The operators will reconcile only their respective cluster types respectively (ExternalCluster or GardenerCluster).

I believe it can be dropped at all - for now, there is no use case where the common API would be required (AFAIK). It can be easily worked around, the scripts and other potential automation would just need to account for both types of clusters. Perhaps @pbochynski or @ebensom may have more insights on the requirements here.

concepts/infrastructure-reconciliation/README.md
janmedrek: Yeah, was thinking about the name for so long I've skipped the best (and simple) option. 😄 Will update the CR name to the `ExternalCluster`.

concepts/infrastructure-reconciliation/README.md
janmedrek: Yeah, that was my observation as well. On the other hand - there won't be that many operators that require those privileges. In addition, all of the affected ones will be located in the Control Plane cluster (Lifecycle Manager, Compass, ...?), and the rest will run in the SKR cluster. I think we can reevaluate this matter with the security guys as quite a few things changed since the last discussion.

concepts/infrastructure-reconciliation/README.md
jakobmoellersap: I think Option 1 (the Watch) we can only consider if we can guarantee that the Secret Resource Versions are not updated in second-intervals. (I think this is unrealistic in a real cluster as there are lots of interactions for e.g. creating annotations, events, etc. that could influence the resource version of Secrets). But looking at the alternatives, we should at least try it out in a synthetic load test before jumping to conclusions early
 
The second option would only work if the timestamp would allow us to limit our Watch Requests (meaning before we are caching in the controller and doing the request against the API Server), with something like a Label/Field Selector. I think this is currently impossible as Label Selectors are limited to Equality Comparisons (https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors) and Field Selectors do not work on Secret Fields (https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/). This would only leave us with Annotations, but these are not supported for filtering in the Watch API of Kubernetes (only label selectors are supported here for server side filtering). I still think your idea could work though if we find a proper filtering criteria. 

Regarding your third option: I think realistically speaking we can use the "dumb" retry method. It will most likely lead to a solution of workers creating clients on-demand with a certain lifetime in a pool. In case we then discover they fail, we can just error out, kill the client from the pool and then retry in a second reconcile loop downstream (e.g. in lifecycle manager for synchronizing the Kyma CR). We can constrain memory consumption by only holding clients for a limited amount of clusters (with the pool being controlled by MFU for requests) and evicting clients that are not reconciled often and recreate them on demand.

concepts/infrastructure-reconciliation/README.md
pbochynski: I agree - we can postpone creating generic Cluster CR when we have clear demand for it.

pbochynski (approved):

concepts/infrastructure-reconciliation/README.md
janmedrek: It quite depends on the validity time of kubeconfigs we create, I suppose it will not be measured in hours - rather days. Then, in the "retry" case we would expect one retry per cluster, per two or three days, which in my opinion, even with a big number of clusters should not influence the performance and the resource usage in any meaningful way.

Regarding the timestamps, I've rather thought about implementing the refetch logic in the operators using those kubeconfigs rather than limiting their Watch Requests. This way the timestamp would be cached with the kubeconfig and the operator itself would determine whether the kubeconfig is outdated and it needs a new one or can use the one from the cache. This has some drawbacks though. It would require such a mechanism to be implemented across multiple different components and I am not sure caching the kubeconfigs for all the clusters in the KCP would be feasible.

jakobmoellersap: The Watch is not a request per update of `data` in a secret, but rather one request that streams responses down best-effort and is patched together in the informer based on `resourceVersion`. That means that even if the retry is not that frequent, resource version changes and abortions of the Stream Request will still lead to lots of requests and informer cache updates. (e.g. a simple annotation change from any operator will lead to informer updates and rereads and invalidate the secret containing the kubeconfig).

From what I understand, the cache you are talking about would be fully custom and not use any informer cache, I am strongly opinionated on not writing any custom caching.

I strongly suggest we star with load tests for this early on and choose the approach based on the observations. Otherwise I think we have not enough data to do a good decision here.

janmedrek: Sure, that is PoC material, we can leave that part open for now, I will describe the options in the proposal. 🙂

jakobmoellersap (dismissed):

concepts/infrastructure-reconciliation/README.md
```diff
@@ -1,41 +1,172 @@
...
+		  verbs: ["*"]
+```
+
+This is the simpler one, as there is no cluster provisioning happening in this flow most of the parameters can be dropped.
+
+### Issuing kubeconfigs
+
+As stated before, the separation of kubeconfigs and limiting access is needed to have a healthy, secure environment. Cluster Managers will provide a possibility to create RBAC-based kubeconfigs in both BYOC and Gardener clusters. This will provide operators with the required access to perform any necessary actions.
+
+The kubeconfigs for specific Kubernetes clusters will be fetched (both from Vault or Gardener) in the runtime and stored only in the components' memory for the limited time needed to issue another (constrained) kubeconfig for a given cluster.
```
ebensom, line 135: Gardener Cluster Manager must also need to take care of rotating each shoot's master kubeconfig. I see two options:

1. Keep using the static admin token kubeconfigs for each shoot, and implement configurable time based rotation for them as part of the `GardenerCluster` reconciliation following [this guide](https://github.com/gardener/gardener/blob/master/docs/usage/shoot_credentials_rotation.md#kubeconfig) 
2. Leverate [shoots/adminkubeconfig](https://github.com/gardener/gardener/blob/master/docs/usage/shoot_access.md#shootsadminkubeconfig-subresource) subresource to obtain a  short-living kubeconfig having cluster-admin privileges, when needed (e.g. at the beginning of a `GardenerCluster` reconciliation, when work to be done is detected). Make sure to never persist this kubeconfig in any way, i.e. put to Secrets on KCP.

janmedrek: @ebensom I am wondering, why shouldn't we put that short-living admin kubeconfig into the KCP secret? The issue is that the Gardener Cluster Manager will not be able to determine whether there is any work related to the Kyma installation to be done, thus we won't know when to issue such a kubeconfig.

I would rather say using the shoots/adminkubeconfig subresource is better than having the static ones as we will not need to worry about the rotation, we will just need to make sure there are new ones issued accordingly so that the continuity of operations is preserved. And in case this somehow leaks, it will be invalidated with time, not with a rotation.

janmedrek: This is the issue, to be honest, I am not sure whether the Roles required by the Module Operator will differ vastly from the admin kubeconfig. In that case, we would create an effective duplicate of the original kubeconfig, and in the end, is there a significant difference between using a newly issued one (with admin-like privileges) and the one from the Gardener?

Since the security assessment, the design has changed a bit. There will be no 3rd party components using those kubeconfigs, and the operators for the specific modules were migrated to the SKR. This leaves only the Module Manager and Lifecycle Manager using those. For the Compass Manager, we will have a separate RBAC kubeconfig for sure though.

concepts/infrastructure-reconciliation/README.md
```diff
@@ -1,41 +1,183 @@
-# Reconciler infrastructure changes proposal
+# Cluster Manager proposal
 
 ## Motivation
-Reconciler is a framework to install, update, and repair Kyma components in managed Kyma Runtimes. The reconciliation runs in the loop to ensure that all components are up and running and properly configured. This principle can be extended to cover also infrastructure components, such as virtual networks (VNet) or Kubernetes clusters, but requires some adjustments in the Reconciler architecture. 
+Operators use Kubernetes' Custom Resources as an API and introduce a control loop to manage third party applications and their components. The resources are reconciled periodically to ensure that all desired components are configured properly and running. This principle can be applied to cover also infrastructure components, this will introduce a lower maintenance effort and provides an easy to use, declarative approach to infrastructure provisioning.
```
grego952, line 4:
```suggestion
Operators use Kubernetes' custom resources (CR) as an API and introduce a control loop to manage third-party applications and their components. The resources are reconciled periodically to ensure that all desired components are configured properly and running. This principle can be applied to cover also infrastructure components. This introduces a lower maintenance effort and provides an easy-to-use, declarative approach to infrastructure provisioning.
```

concepts/infrastructure-reconciliation/README.md
```diff
@@ -1,41 +1,183 @@
...
 
-## New use cases for Reconciler
+## Use cases for Cluster Manager
 
-- Create VNet for Kyma Runtime 
-- Create/reconcile a shoot cluster for Kyma Runtime 
-- Create kubeconfig for provisioned clusters (can be rotated)
+- Create/reconcile a Gardener Shoot cluster for Kyma Runtime 
+- Create kubeconfigs for provisioned clusters, those should be limited by Kubernetes roles and rotated regularly. There should be also a possibility to revoke access to the cluster if needed.
+- Manage hyperscaler subscriptions that are used by the Kubernetes Clusters
```
grego952, line 10:
```suggestion
- Manage hyperscaler subscriptions that are used by the Kubernetes clusters
```

concepts/infrastructure-reconciliation/README.md
```diff
@@ -1,41 +1,183 @@
...
-- Register Kyma Runtime within the CMP
+- Trigger reconciliation on a configuration or a cluster state change
 
 ## Proposed changes
 
-### Cluster Inventory as a separate service
+### Cluster Managers handling cluster provisioning
 
-Cluster Inventory is the internal component responsible for reconciliation persistence. It keeps cluster information, configuration, and reconciliation status. Component Reconcilers do not have access to the inventory and cannot modify it directly. One option would be to extend the action context passed to the Reconciler to include more data, but the better option would be to pass only information on how to access Cluster Inventory entry (URL/ID) and allow Components Reconciler to query and update it.
+To achieve a declarative and responsive way of interacting with the infrastructure provisioning a pair of Cluster Managers should be introduced. This components will use an event-based reconciliation of a new Cluster Custom Resources, which will serve as an API for Kubernetes Cluster provisioning.
```
grego952, line 18:
```suggestion
To achieve a declarative and responsive way of interacting with the infrastructure provisioning, a pair of Cluster Managers should be introduced. These components will use an event-based reconciliation of new cluster CRs, serving as an API for Kubernetes Cluster provisioning.
```

concepts/infrastructure-reconciliation/README.md
```diff
@@ -1,41 +1,183 @@
...
 
-### Reconciliation based on watchers and triggers
-Reconciliation is currently time-based (triggered in regular intervals). It is a known limitation. In some cases, the reconciliation should be triggered immediately after some events, such as:
-- Configuration change (e.g. Network Reconciler can add a VNet name to the Cluster Inventory, which would trigger Shoot Reconciler)
-- Resource in the cluster modified (e.g. Connectivity Service instance created in the cluster would trigger Connectivity Proxy Reconciler)
+In order to provide a common Custom Resource for all clusters regardless of their licence type there will be a new pair of cluster CRs introduced:
```
grego952, line 20:
```suggestion
In order to provide a common CR for all clusters regardless of their license type there will be a new pair of cluster CRs introduced:
```

concepts/infrastructure-reconciliation/README.md
```diff
@@ -1,41 +1,183 @@
...
+- GardenerCluster CR (for managed clusters running on Gardener)
+- ExternalCluster CR (for Bring-Your-Own-Cluster model)
 
-Restoring managed resources accidentally (or intentionally) deleted by a user is not the main goal of watchers. They should be rather used to monitor resources such as service instances that should trigger some integration tasks in the Kyma Control Plane.
+And respectively - there will be two operators handling different processes for those CRs:
+- Gardener Cluster Manager, which will handle Gardener Shoot provisioning and kubeconfig generation/rotation
```
grego952, line 25:
```suggestion
- Gardener Cluster Manager, which handles Gardener Shoot provisioning and kubeconfig generation/rotation
```

concepts/infrastructure-reconciliation/README.md
```diff
@@ -1,41 +1,183 @@
...
+
+### Handling hyperscaler subscriptions
+
+Separating user subscriptions is both a business and security requirement. Right now it is handled by the Hyperscaler Account Pool package in the Kyma Environment Broker, which is not a good place to have that. The proposal is to provide an operator that would handle the available hyperscaler subscription pool, deciding whether Gardener's Shoot should be provisioned with the already used credential set or whether a new one is needed. This introduces an abstraction layer which separates both Kyma Environment Broker and Cluster Manager from the subscription handling logic and provides a way to easily extend that system in the future.
+
+### Issuing RBAC-based kubeconfigs
+
+To provide an environment, where certain Operators are not able to leverage the cluster access, it is required to open up a possibility for the constrained kubeconfigs to be created. This ensures that Operators can work only with the resources they require access to and security-wise, potential malicious parties won't be able to use kubeconfigs issued for the Operators to execute cluster-wide attacks.
```
grego952, line 42:
```suggestion
To provide an environment where certain Operators cannot leverage the cluster access, it is required to open up a possibility for the constrained kubeconfigs to be created. This ensures that Operators can work only with the resources they need access to. Security-wise, potential malicious parties won't be able to use kubeconfigs issued for the Operators to execute cluster-wide attacks.
```

concepts/infrastructure-reconciliation/README.md
```diff
@@ -1,41 +1,183 @@
...
 
 ![](assets/infrastructure-reconciliation.svg)
 
-Mothership Reconciler should provide common functionality related to watching changes and triggering Component Reconcilers in response to events or with the time-based schedule. Such aspects as queueing reconciliation requests, consolidating them, updating reconciliation status, retries and back-off strategy should be implemented once (in the Mothership). 
+The three new operators would need to be introduced:
+- Gardener Cluster Manager
+- External Cluster Manager
+- Subscription Manager
+
+Each of those components has a clear set of responsibilities that together handle the flow that creates a new, complete, Kyma-ready  Kubernetes cluster.
```
grego952, line 59:
```suggestion
Each of those components has a clear set of responsibilities that together handle the flow that creates a new, complete, Kyma-ready Kubernetes cluster.
```

concepts/infrastructure-reconciliation/README.md
```diff
@@ -1,41 +1,183 @@
...
+
+Along the operators, there are four new Custom Resources, that serve as an API for the components:
```
grego952, line 61:
```suggestion
Along the operators, there are four new CRs that serve as an API for the components:
```

concepts/infrastructure-reconciliation/README.md
```diff
@@ -1,41 +1,183 @@
...
+- Gardener Cluster CR
+- External Cluster CR
+- SubscriptionRequest CR
+
+### Cluster Managers and the Cluster CRs
+
+The Cluster Managers' main responsibility is to manage the respective Cluster CRs, this is tied to both their provisioning and providing an access to the created K8S clusters.
```
grego952, line 68:
```suggestion
The Cluster Managers' main responsibility is to manage the respective Cluster CRs, this is tied to both their provisioning and providing an access to the created Kubernetes clusters.
```

concepts/infrastructure-reconciliation/README.md
```diff
@@ -1,41 +1,183 @@
...
+		rules:
+		- apiGroups: ["*"]
+		  resources: ["*"]
+		  verbs: ["*"]
+	administrators: ["user.id@sap.com", "other.id@sap.com"]
+```
+
+This contains a minimal set of values that are required to provision a new Gardener cluster. The resource also contains the `administrators` and `subscriptionSelector` fields that are required for the Binding creation and requesting a hyperscaler subscription respectively.
+
+And the second one for BYOC model:
```
grego952, line 111:
```suggestion
And the second one for the Bring Your Own Cluster (BYOC) model:
```

concepts/infrastructure-reconciliation/README.md
```diff
@@ -1,41 +1,183 @@
...
+    name: subscription-request
+    namespace: cluster-manager
+spec:
+    provider: aws
+    subscriptionSelector: "abcd-0001"
+status:
+    gardenerSecretName: subscription-secret
+```
+
+The resource structure is really simple, the Subscription Manager only needs to know the provider and a proper selector for the subscription to be selected. It provides credentials in a form of the Gardener secret name, which can be simply used as a parameter in the Shoot CR.
```
grego952, line 153:
```suggestion
The resource structure is really simple. Subscription Manager only needs to know the provider and a proper selector for the subscription to be selected. It provides credentials in the form of the Gardener's Secret name, which can be simply used as a parameter in the Shoot CR.
```

concepts/infrastructure-reconciliation/README.md
```diff
@@ -1,41 +1,183 @@
...
+
+The flow from the user's perspective stays intact, the part that changes is the list of actions needed by KEB to make sure the cluster is up and running.
+
+First of all, after receiving the provisioning request, KEB should create a set of CRs for accordingly to the licence type of the Runtime:
+- For BYOC - ExternalCluster CR
+- For managed - GardenerCluster CR
+
+KEB should deliver all the required parameters for the cluster provisioning, this translates to the minimal infrastructure config for Gardener clusters, and the role/kubeconfig details for both types.
+
+Another thing that KEB would need to create is the Kyma CR for the Runtime, this will either require passing the proper secret reference so that kubeconfig can be accessed by the Lifecycle Manager.
```
grego952, line 165:
```suggestion
Another thing that KEB would need to create is the Kyma CR for the Runtime, this will either require passing the proper Secret reference so that kubeconfig can be accessed by the Lifecycle Manager.
```

concepts/infrastructure-reconciliation/README.md
```diff
@@ -1,41 +1,183 @@
...
+After the CRs are created, the respective Cluster Manager will pick up the provisioning request and will create a new Shoot for the runtime (managed scenario only), and provide a requested kubeconfig.
+
+At this point Lifecycle Manager can step in, it has all required resources to set up the Runtime:
+- running cluster
+- kubeconfig (passed via secret reference)
+- Kyma CR
+
+## Reacting to the kubeconfig rotation
+
+Since Lifecycle Manager and Module Manager need to interact with the provisioned K8S cluster directly, a mechanism for providing them with an up-to-date kubeconfig is needed.
```
grego952, line 176:
```suggestion
Since Lifecycle Manager and Module Manager need to interact with the provisioned Kubernetes cluster directly, a mechanism for providing them with an up-to-date kubeconfig is needed.
```

concepts/infrastructure-reconciliation/README.md
```diff
@@ -1,41 +1,183 @@
...
 
-Worker pool implementation has to be revisited. Mothership should not manage dependencies - Component Reconciler should be written in a way that can recognize missing dependencies and fail fast. The next reconciliation invocation should happen when the configuration (Cluster Inventory) is updated (dependency). 
+There are few possible options here:
+- Watching K8S Secrets - Managers should watch the Secrets containing the kubeconfigs, with a big amount of Runtimes this may be resource intensive on the API server side.
```
grego952, line 179:
```suggestion
- Watching Kubernetes Secrets - Managers should watch the Secrets containing the kubeconfigs, with a big amount of Runtimes this may be resource intensive on the API server side.
```

concepts/infrastructure-reconciliation/README.md
ebensom: In that case, no. there isn't much of a difference in terms of granted permissions.
The Gardener issued kubeconfig uses client certificate authentication, where the certificate has the short expiration date, and the identity of the certificate (DN) is the name of the service account which requested the kubeconfig (e.g. compass-provisioner).
In the case of the proposal's RBAC kubeconfig, it uses token issued for a K8S service account created on the cluster by the cluster operator, which would be expired and deleted from the cluster after some time.
Ultimately they all grant cluster-admin permissions to a technical user identity, so no big difference from result point of view.

grego952 (approved):

pbochynski (approved):

PR #727, opened 2022-10-18
Support compatibility with Pod Security Standards

TorstenD-SAP:
Documentation and Pod Security Standards compliant Pod specification template

**Description**

- Introduce a documentation about the replacement of Pod Security Policies and the usage of the Pod specification template
- Introduce Pod specification templates containing security-related settings which are compliant with the Pod Security Standards

**Related issue(s)**


concepts/psp-replacement/restricted-pod-spec.yaml
```diff
@@ -0,0 +1,85 @@
...
+      - name: net.ipv4.ip_unprivileged_port_start,
+        value: ''
+      - name: net.ipv4.tcp_syncookies,
+        value: ''
+      - name: net.ipv4.ping_group_range,
+        value: ''
+  hostNetwork: false # Optional. The default is false if the entry is not there.
+  hostPID: false # Optional. The default is false if the entry is not there.
+  hostIPC: false # Optional. The default is false if the entry is not there.
+  volumes: # Only the volume types in the following list are allowed. If no volumes needed, the whole volumes-section has to be removed. Otherwise replace the brackets with the volumes needed.
```
IwonaLanger, line 31:
```suggestion
  volumes: # Only the volume types in the following list are allowed. If no volumes needed, the whole volumes-section has to be removed. Otherwise, replace the brackets with the volumes needed.
```

concepts/psp-replacement/restricted-pod-spec.yaml
```diff
@@ -0,0 +1,85 @@
...
+        - containerPort: <integer> # Number of port to expose on the pod's IP address. This must be a valid port number, 0 < x < 65536.
+          hostPort: 0 # 0 if no host port is needed. Specify a dedicated one if necessary.
+      securityContext:
+        privileged: false # Mandatory
+        allowPrivilegeEscalation: false # Mandatory
+        runAsNonRoot: true # Mandatory
+        capabilities:
+          drop: # Mandatory.
+            - ALL
+          add: # Only the NET_BIND_SERVICE capability is allowed in the restricted standard. If it is not needed, the whole add-section has to be removed
```
IwonaLanger, line 62:
```suggestion
          add: # Only the NET_BIND_SERVICE capability is allowed in the restricted standard. If it is not needed, the whole add-section has to be removed.
```

concepts/psp-replacement/restricted-pod-spec.yaml
```diff
@@ -0,0 +1,85 @@
...
+  initContainers:
+    - name: init-container-restricted
+      securityContext:
+        privileged: false # Mandatory
+        allowPrivilegeEscalation: false # Mandatory
+        runAsNonRoot: true # Mandatory
+        capabilities:
+          drop: # Mandatory
+            - ALL
+          add: # Only the NET_BIND_SERVICE capability is allowed in the restricted standard. If it is not needed, the whole add-section has to be removed
```
IwonaLanger, line 77:
```suggestion
          add: # Only the NET_BIND_SERVICE capability is allowed in the restricted standard. If it is not needed, the whole add-section must be removed.
```

concepts/psp-replacement/restricted-pod-spec.yaml
```diff
@@ -0,0 +1,85 @@
...
+            - NET_BIND_SERVICE
+        seLinuxOptions: # Set this on container level and only if it is really necessary. Setting the user or role option is forbidden. If the setting is not needed, the whole seLinuxOptions-section has to be removed.
+          type: container_t # or container_init_t or container_kvm_t
+        procMount: default
+        readOnlyRootFilesystem: true # Mandatory, Security Best Practice. Not relevant for Pod Security Admission
+      ports: # Delete section if it is not needed
```
IwonaLanger, line 83:
```suggestion
      ports: # Delete section if it is not needed.
```

concepts/psp-replacement/restricted-pod-spec.yaml
```diff
@@ -0,0 +1,85 @@
...
+        - containerPort: <integer> # Number of port to expose on the pod's IP address. This must be a valid port number, 0 < x < 65536.
+          hostPort: 0 # 0 if no host port is needed. Specify a dedicated one if necessary
```
IwonaLanger, line 85:
```suggestion
          hostPort: 0 # 0 if no host port is needed. Specify a dedicated one if necessary.
```

concepts/psp-replacement/README.md
```diff
@@ -0,0 +1,31 @@
...
+| Seccomp                  | Optional<br><br>Allowed values are:<br>- RuntimeDefault<br>- Localhost                                                                                                                                                                                                  | Mandatory<br><br>Allowed values are:<br>- RuntimeDefault<br>- Localhost                                                                                                                                                                      |
+| Capabilities             | Only additional capabilities from the list below are allowed.<br><br>- Undefined/nil<br>- AUDIT_WRITE<br>- CHOWN<br>- DAC_OVERRIDE<br>- FOWNER<br>- FSETID<br>- KILL<br>- MKNOD<br>- NET_BIND_SERVICE<br>- SETFCAP<br>- SETGID<br>- SETPCAP<br>- SETUID<br>- SYS_CHROOT | Containers must drop ALL capabilities, and are only permitted to add back the NET_BIND_SERVICE capability.                                                                                                                                   |
```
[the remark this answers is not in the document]
TorstenD-SAP: I change the value of the 'Running as Non-root user' setting a bit but I didn't follow your suggestion because the developer can set it to 0 and there might be some circumstances where this is necessary.

TorstenD-SAP (commented):
Adopt the suggestions from the review

concepts/psp-replacement/README.md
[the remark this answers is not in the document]
IwonaLanger: I didn't intend to change the meaning here, I simply applied the content guidelines. For the sake of language simplicity, we are not supposed to use passive voice forms. "Should" is also to be avoided in favor of "must" or "can". Hence the change from "should never" to "cannot". If you insist on keeping the original sentence, just change the form of "be" in "it should never been set" (from "been" to "be"), please.

pbochynski (commented):
Nice document, explains a lot, but I miss practical guide. 
- What should I do as a regular Kyma developer to verify if my deployments are compliant? 
- Should I be compliant with baseline or with restricted? 
- What if my deployment is not compliant but I can't make it?

TorstenD-SAP:
@pbochynski I added some actions to give more guidance to developers. In addition, I will create Github issues referencing this guide and the templates in order to get this done.

pbochynski (approved):

PR #736, opened 2022-11-07
Modularization concept update

pbochynski:
**Description**

- more examples of release channels in the modularization concept
- principles for managing component dependencies
- release flow from team perspective

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/13586


concepts/modularization/README.md
```diff
@@ -41,13 +41,47 @@ If the API you need (like a core Kubernetes API or Istio virtual service) is not
 # Release channels
 Release channels let customers try new modules and features early, and decide when the updates should be applied. 
 
+The first use case will be modeled as the alpha channel. Modules available in the alpha channel are developed with all quality measures in place (functional correctness, security, etc.), but they might still have unstable API or be changed without keeping backward compatibility. When you use a module from the alpha channel, you won't get full SLA guarantees for that module or other modules that are affected (directly or indirectly).
```
jakobmoellersap, line 44: I think we should add that dependencies can break and there is no guarantee on the removal of a module at a later point (even if its already implicitly said as its not part of SLAs)

concepts/modularization/README.md
```diff
@@ -41,13 +41,47 @@ If the API you need (like a core Kubernetes API or Istio virtual service) is not
...
+
+The second use case (deciding when updates should be applied) will require 2 production-grade channels with a different update schedule. The fast channel will get updates as soon as they are released and have passed all quality gates. The regular channel will get updates a few days later. Customers can switch the entire cluster or a particular component to the regular channel to check if the upstream changes do not cause any issues with their workload. Changing back to the regular channel is possible, but the module version won't be downgraded - the next version has to reach the channel to be applied.
+
+Moving module versions between channels should be automated as much as possible using a promotion strategy suitable for the particular version change (major/minor/patch) or priority (regular/hot-fix).
```
jakobmoellersap, line 48: QQ: isnt semver already covering both of these? I think we should just reference Semver as the only valid way to increment and introduce patches/fixes/features

concepts/modularization/README.md
```diff
@@ -41,13 +41,47 @@ If the API you need (like a core Kubernetes API or Istio virtual service) is not
...
+In the diagram, we have an example with 3 modules that got new releases. Each one represents a different promotion strategy:
+
+**Module X**
+
+Module `x` is in active development. Version 1.0.x is out for a longer time (version 1.0.18 in the regular channel). The new version (1.1.0) was released recently and is available in the fast channel for a few days. The new version is not super stable yet. The patch version (1.1.1) with bug fixes was required a few days later. Also, the security vulnerability was fixed in that patch but the version is too fresh to go to the regular channel already, therefore new patch release for the 1.0.x version is required (only the security patch included). The patch goes only to the regular channel as we can't downgrade versions in the alpha or fast channel.
```
jakobmoellersap, line 56: This patching to the regular channel is already an exception to our timed flow of 2 weeks incubation in the fast channel. I think we should judge carefully here since this would mean we have to introduce release branches.

concepts/modularization/README.md
```diff
@@ -41,13 +41,47 @@ If the API you need (like a core Kubernetes API or Istio virtual service) is not
...
+
+**Module X**
+
+Module `x` is in active development. Version 1.0.x is out for a longer time (version 1.0.18 in the regular channel). The new version (1.1.0) was released recently and is available in the fast channel for a few days. The new version is not super stable yet. The patch version (1.1.1) with bug fixes was required a few days later. Also, the security vulnerability was fixed in that patch but the version is too fresh to go to the regular channel already, therefore new patch release for the 1.0.x version is required (only the security patch included). The patch goes only to the regular channel as we can't downgrade versions in the alpha or fast channel. 
+
+The team develops new features and soon introduces another minor version (1.2.0). Shipping that version to the fast channel can cause problems as we already have 2 older versions in the active maintenance. We can either decide to push it to the fast channel (replacing 1.1.x) immediately or wait until 1.1.x is promoted to the regular first. The first case extends the maintenance period for 1.0.x version and requires direct upgrade from 1.0.x to 1.2.x. The second case means 1.2.x availability has to be delayed.
```
jakobmoellersap, line 58:
```suggestion
The team develops new features and soon introduces another minor version (1.2.0). Shipping that version to the fast channel can cause problems as we already have 2 older versions in the active maintenance. We can either decide to push it to the fast channel (replacing 1.1.x) immediately or wait until 1.1.x is promoted to the regular first. The first case extends the maintenance period for 1.0.x version and requires direct upgrade from 1.0.x to 1.2.x. The second case means 1.2.x availability has to be delayed. 
```

Why would the upgrade cause problems if you are patching from a patch that is not the most recent to the next minor?
I would say in this case, it does not extend the maintenance period. I agree with the direct upgrade requirement however.
For the second case, I do not understand why availability is delayed. Do you mean the 2 weeks duration until its available in regular again?

concepts/modularization/README.md
```diff
@@ -41,13 +41,47 @@ If the API you need (like a core Kubernetes API or Istio virtual service) is not
 # Release channels
 Release channels let customers try new modules and features early, and decide when the updates should be applied. 
 
+The first use case will be modeled as the alpha channel. Modules available in the alpha channel are developed with all quality measures in place (functional correctness, security, etc.), but they might still have unstable API or be changed without keeping backward compatibility. When you use a module from the alpha channel, you won't get full SLA guarantees for that module or other modules that are affected (directly or indirectly).
+
+The second use case (deciding when updates should be applied) will require 2 production-grade channels with a different update schedule. The fast channel will get updates as soon as they are released and have passed all quality gates. The regular channel will get updates a few days later. Customers can switch the entire cluster or a particular component to the regular channel to check if the upstream changes do not cause any issues with their workload. Changing back to the regular channel is possible, but the module version won't be downgraded - the next version has to reach the channel to be applied.
```
janmedrek, line 46: I believe you've meant that the whole cluster can be switched to the fast channel (to test the new version), a small typo:
```suggestion
The second use case (deciding when updates should be applied) will require 2 production-grade channels with a different update schedule. The fast channel will get updates as soon as they are released and have passed all quality gates. The regular channel will get updates a few days later. Customers can switch the entire cluster or a particular component to the fast channel to check if the upstream changes do not cause any issues with their workload. Changing back to the regular channel is possible, but the module version won't be downgraded - the next version has to reach the channel to be applied.
```

concepts/modularization/README.md
```diff
@@ -41,13 +41,49 @@ If the API you need (like a core Kubernetes API or Istio virtual service) is not
...
 
 The second use case (deciding when updates should be applied) will require 2 production-grade channels with a different update schedule. The fast channel will get updates as soon as they are released and have passed all quality gates. The regular channel will get updates a few days later. Customers can switch the entire cluster or a particular component to the fast channel to check if the upstream changes do not cause any issues with their workload. Changing back to the regular channel is possible, but the module version won't be downgraded - the next version has to reach the channel to be applied.
 
-Hotfixes will be delivered to all channels immediately (TODO: how to apply a hotfix for the release that is not available in the current channel).
+Moving module versions between channels should be automated as much as possible using a promotion strategy suitable for the particular version change (major/minor/patch) or priority (regular/hot-fix).
```
pbochynski, line 48:
```suggestion
Moving module versions between channels should be automated as much as possible using a promotion strategy suitable for the particular version change (major/minor/patch defined by [semver.org](https://semver.org/)) or priority (regular/hot-fix).
```

concepts/modularization/README.md
```diff
@@ -41,13 +41,49 @@ If the API you need (like a core Kubernetes API or Istio virtual service) is not
...
+
+**Module X**
+
+Module `x` is in active development. Version 1.0.x is out for a longer time (version 1.0.18 in the regular channel). The new version (1.1.0) was released recently and is available in the fast channel for a few days. The new version is not super stable yet. The patch version (1.1.1) with bug fixes was required a few days later. Also, the security vulnerability was fixed in that patch but the version is too fresh to go to the regular channel already, therefore new patch release for the 1.0.x version is required (only the security patch included). The patch goes only to the regular channel as we can't downgrade versions in the alpha or fast channel. 
+
+The team develops new features and soon introduces another minor version (1.2.0). Shipping that version to the fast channel can cause problems as we already have 2 older versions in the active maintenance. We can either decide to push it to the fast channel (replacing 1.1.x) immediately or wait until 1.1.x is promoted to the regular first. The first case extends the maintenance period for 1.0.x version and requires direct upgrade from 1.0.x to 1.2.x. The second case means 1.2.x availability has to be delayed.
```
pbochynski, line 58:
```suggestion
The team develops new features and soon introduces another minor version (1.2.0). Shipping that version to the fast channel can cause problems as we already have 2 older minor versions in active maintenance. We can either decide to push it to the fast channel (replacing 1.1.x) immediately or wait until 1.1.x is promoted to the regular first. The first case extends the maintenance period for 1.0.x version (2 weeks waiting period is reset) and requires a direct upgrade from 1.0.x to 1.2.x. The second case means 1.2.x availability has to be delayed (it can be introduced after 1.0.x is replaced by 1.1.x in the regular channel). 
```

concepts/modularization/README.md
pbochynski: We follow semantic versioning, but it is more about the delays we introduce to increase stability. Nevertheless, I added link to semver.org

concepts/modularization/README.md
pbochynski: Release branches are there and are inevitable. It is more important to use them only for hot-fixes, not for regular development. If we do not have anything critical the next version from main branch should bring bug fixes and security patches.

concepts/modularization/README.md
pbochynski: If we upgrade the fast channel to 1.2.0 the stabilization time is reset (2 weeks until promotion). Then the 1.0 version is available for a little bit longer time.
If we don't upgrade the fast channel to 1.2 before 1.1 goes to regular we delay its promotion.

concepts/modularization/README.md
```diff
@@ -41,13 +41,49 @@ If the API you need (like a core Kubernetes API or Istio virtual service) is not
 # Release channels
 Release channels let customers try new modules and features early, and decide when the updates should be applied. 
 
-![](assets/release-channels.drawio.svg)
-
 The first use case will be modeled as the alpha channel. Modules available in the alpha channel are developed with all quality measures in place (functional correctness, security, etc.), but they might still have unstable API or be changed without keeping backward compatibility. When you use a module from the alpha channel, you won't get full SLA guarantees for that module or other modules that are affected (directly or indirectly).
```
tobiscr, line 44:
```suggestion
The first use case will be modeled as the alpha channel. Modules available in the alpha channel are developed with all quality measures in place (functional correctness, security, etc.), but they might still have unstable API, be changed without keeping backward compatibility or even removed before they reach fast or regular channel. When you use a module from the alpha channel, you won't get full SLA guarantees for that module or other modules that are affected (directly or indirectly).
```

concepts/modularization/README.md
```diff
@@ -38,17 +39,76 @@ Components can depend only on core Kubernetes API, or on API extensions introduc
...
 
 The second use case (deciding when updates should be applied) will require 2 production-grade channels with a different update schedule. The fast channel will get updates as soon as they are released and have passed all quality gates. The regular channel will get updates a few days later. Customers can switch the entire cluster or a particular component to the fast channel to check if the upstream changes do not cause any issues with their workload. Changing back to the regular channel is possible, but the module version won't be downgraded - the next version has to reach the channel to be applied.
 
-Hotfixes will be delivered to all channels immediately (TODO: how to apply a hotfix for the release that is not available in the current channel).
+Moving module versions between channels should be automated as much as possible using a promotion strategy suitable for the particular version change (major/minor/patch defined by [semver.org](https://semver.org/)) or priority (regular/hot-fix).
```
janmedrek, line 56: Regarding this one, not a change request but rather a clarification and a sanity check on how it should be approached in the new architecture we are currently working on.

In my opinion, the promotion topic should be isolated from the Lifecycle Manager's logic. This will provide better separation and result in a clean cut between the responsibilities of the components. Therefore, my proposed solution would be:
- Have the SemVer kept for the Modules
- Make the Lifecycle Manager a bit channel-agnostic (we won't separate from that completely as channels are an essential part of current API implementation)

As for the result, Lifecycle Manager will [no longer use an enum for the release channel matching](https://github.com/kyma-project/lifecycle-manager/issues/324), it will just do the generic string comparison. This will provide a way of creating ad-hoc testing release channels in the dev environments (if needed) that can be used for some extreme-case upgrade scenarios that have to be tested (and by the way it will secure the implementation from any changes in the naming of the channels).

Then, the Lifecycle Manager will have 3 "primitive" scenarios to handle - installation, upgrade and downgrade. SemVer of the currently installed and the new Module would be used for determining if any action is needed, and as agreed - downgrades will not be supported.

And then - any automation taking care of the promotion logic of the modules can be implemented independently.

pbochynski:
```suggestion
Moving module versions between channels can be done with a simple pull request (bump version). Later on, it can be automated using a promotion strategy suitable for the particular version change (major/minor/patch defined by [semver.org](https://semver.org/)) or priority (regular/hot-fix), but it is not the highest priority for the first iteration.
```

pbochynski: Yes - I agree. Promotion logic is not part of the lifecycle manager - it is CI/CD automation.

janmedrek (approved):

PR #755, opened 2023-01-10
Modularization - migration process

pbochynski:
**Description**

- migration plan for modularization - switch from reconciler to lifecycle-manager

**Related issue(s)**

### DIFF
--- a/concepts/modularization/migration.md
+++ b/concepts/modularization/migration.md
@@ -1,0 +1,55 @@
+# Migration plan
+
+Migration plan will be executed several times until all modules are moved from reconciler to lifecycle-manager. Is it possible to migrate several modules at once but it is not recommended. The migration plan uses `serverless` module as an example.
+
+## Prerequisite: enable lifecycle-manager
+- KEB creates Kyma resources in the control plane. Lifecycle-manager is deployed in KCP and enabled for all clusters with empty module list.
+
+## Enable module for new clusters (first)
+
+First step is to enable module installation with lifecycle manager for new clusters. Existing clusters are still managed by reconciler.
+- Serverless operator(manager) is ready and module template is available in the fast/regular channel. 
+- New Kyma version is enabled in KEB (without serverless module). New clusters do not have serverless until customer enabled it in Kyma resource.
+
+
+## Update existing clusters
+
+Migration can be stopped at any moment if any problem occurs. Not migrated clusters will still use serverless reconciler (module not enabled in KCP)
+
+1. Configure reconciler to ignore serverless component is the module is enabled in Kyma CR (KCP or SKR).
+
+2. Migration script will be executed for existing clusters (in several groups - not all cluster at once). The migration script actions:
+- create CRD and default resource for serverless manager in the SKR (we can have different CRs for different plans/profiles if needed)
+- add module to kyma resource in KCP
+
+3. When all clusters are migrated run migration finalization script:
+- add module to Kyma CR in SKR
+- remove serverless module from the Kyma CRs in the KCP 
+
+
+![migration](assets/migration.drawio.svg)
+
+
+## Finalization - post migration steps
+
+- Upgrade all the clusters to the new Kyma version (without serverless)
+- Remove serverless reconciler from the code.
+
+# Validation scenarios
+
+The migration process should be resilient and validated against possible user actions that could cause problems. It is mainly about turning on and off the module that is being migrated:
+- in the new cluster - OK (only operator works as module is not included in the KEB)
+- in the old (not migrated) cluster before migration script is executed - almost OK (turning on is OK, but turning off will cause reconciler running in parallel with uninstallation from lifecycle manager)
+- in the old cluster during migration - OK (no effect as module is also enabled in SKR)
+- in the old cluster after migration - OK (new version without serverless module is configured in the reconciler)
+
+
+# Challenges
+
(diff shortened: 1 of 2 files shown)


concepts/modularization/migration.md
```diff
@@ -0,0 +1,36 @@
...
+1. Upgrade the clusters to the new Kyma version (without serverless)
+
+2. Create CRD and default resource for serverless manager in the SKR (we can have different CRs for different plans/profiles if needed)
+
+
+![migration](assets/migration.drawio.svg)
+
+# Validation scenarios
+
+The migration process should be resilient and validated against possible user actions that could cause problems. It is mainly about turning on and off the module that is being migrated:
```
jakobmoellersap, line 27: I believe it would be safest to apply the Module in KCP during the migration period with `noModuleCopy` set to `false`. This will cause the module to be reflected in the SKR and even if it is removed by the user, the module will still be active and avoid side effects

concepts/modularization/migration.md
```diff
@@ -0,0 +1,36 @@
...
+- in the new cluster - OK (reconciler is already disabled for serverless)
+- in the old (not migrated) cluster before migration script is executed - almost OK (if user adds serverless module manually it will work in parallel with reconciler)
+- in the old cluster during migration - OK (reconciler is already disabled for serverless)
+- in the old cluster after migration - OK (reconciler is already disabled for serverless)
+
+
+# Challenges
+
+Migration has one challenge: it can become unstable if the user will add and remove module using Kyma CR in SKR when module template is available but before migration has been started. To minimize the risk we can disable module template propagation for old clusters until migration is over.
```
jakobmoellersap, line 36: What do you mean with disable module template propagation for old clusters? I am not sure how this would be achievable. If the ModuleTemplates are available they can be used. I think if we want to disable module enablement we should just disable the Kyma Synchronisation to SKR while migrating and operate solely on KCP. That way we can enable customer modification simply by enabling the sync later on. WDYT?

jakobmoellersap (commented):
Two comments regarding the edge case of customers modifying the Kyma CR before migration is starting. Other than that LGTM 👍

concepts/modularization/migration.md
pbochynski: I was thinking about the `moduleCatalog` property set to false for existing clusters. This way user will not see the module in the UI and will not create it before it is added by the migration script.

concepts/modularization/migration.md
pbochynski: I believe `noModuleCopy` set to `false` means `moduleCopy=true` ;-) (I suggest to avoid negative boolean variable names)
But if we set `noModuleCopy=true` it can work. When users try to enable module on their own it will have no effect. Module CR is not created in the SKR so the module operator is not active yet (assuming the module operator is not doing clean-up if it cannot find any module CR in the cluster)

jakobmoellersap: reason for negative is so that the default bool value can be used and we don't have to write a defaulting web hook to set it to true every time.

I agree with your sentiment on the Module CR. I'll talk with @janmedrek to see if we should prioritise the feature earlier as we originally thought to allow this or if we could go with an alternative.

concepts/modularization/migration.md
```diff
@@ -0,0 +1,32 @@
...
+First step is to enable module installation with lifecycle manager for new clusters. Existing clusters are still managed by reconciler.
+- Disable module template replication for existing clusters (moduleCatalog=false). This prevents users to add serverless module before it is disabled in the reconciler.
+- New Kyma version is enabled in KEB (without serverless module). New clusters do not have serverless until customer enabled it in Kyma resource.
+- Serverless operator(manager) is ready and module template is available in the fast/regular channel. 
+
+## Update existing clusters
+
+Migration can be stopped at any moment if any problem occurs. Not migrated clusters will still use serverless reconciler (old Kyma version with serverless)
+
+- Upgrade the clusters to the new Kyma version (without serverless) - from this point reconciler doesn't touch serverless resources.
```
kwiatekus, line 20: Does this step actually removes installed serverless (workloads) ? or we simply switch off reconciliation os serverless?
 
 It what point (if any ) are the "old serverless" workloads removed?
 I think once we apply Serverless CR we will have duplicated serverless workloads, no?
 
 Or is it expected that serverless operator "takes over" existing installation , i.e by annotating them

concepts/modularization/migration.md
```diff
@@ -0,0 +1,32 @@
...
+- Add module to Kyma CR in SKR with option to not copy default CR. Lifecycle manager will install serverless manager without starting module installation. Default CR creation is omitted for the case when we need different setting for different clusters (eg. trial / production). If default CR can be applied for all the clusters then module copy can stay enabled and the next step is not required.
+- Create CRD  and default CR for serverless manager in the SKR. CRD creation is optional - it can be skipped, but then migration script has to wait until liefecycle-manager installs it. Entire step can be skipped if the default module CR can be applied for all the clusters.
```
ebensom, line 22: At this point, the Kyma CR is expected to be synced back to the KCP, right, so it will reflect that serverless has been enabled on the SKR?

janmedrek (approved):

concepts/modularization/migration.md
pbochynski: To be precise Kyma CR spec from the runtime cluster is not copied back to the control plane, but you can see that serverless is enabled in the status.

concepts/modularization/migration.md
pbochynski: We do not want to remove (uninstall) serverless during migration. Customers expect continuation. We switch off reconciliation and install the serverless operator. The operator has to take ownership of resources created by the reconciler. How it is done - is up to you.

PR #858, opened 2023-11-14
Add action to the whitelist

Cortey:
**Description**
Adds an allowed action used by `Otters` team in the release flow

**Related issue(s)**
https://github.com/kyma-project/serverless/issues/388

### DIFF
--- a/docs/contributing/assets/allowed_actions.json
+++ b/docs/contributing/assets/allowed_actions.json
@@ -140,6 +140,17 @@
         "repository": "https://github.com/suzuki-shunsuke/tfcmt"
       },
       "comment": ""
+    },
+    {
+      "name": "addnab/docker-run-action",
+      "versions": [
+        "v3"
+      ],
+      "repository": "https://github.com/addnab/docker-run-action",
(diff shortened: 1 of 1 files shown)


docs/contributing/assets/allowed_actions.json
```diff
@@ -140,6 +140,17 @@
         "repository": "https://github.com/suzuki-shunsuke/tfcmt"
       },
       "comment": ""
+    },
+    {
+      "name": "addnab/docker-run-action",
+      "versions": [
+        "v3"
```
TorstenD-SAP, line 147:
```suggestion
        "4f65fabd2431ebc8d299f8e5a018d79a769ae185"
```
The approval is bound to a specific version of an action and to bind the action to this version it is necessary to use the commit sha of the version instead of the version tag. Especially if the version tag contains only the major version.

TorstenD-SAP (changes requested):

TorstenD-SAP (changes requested):
See my suggestions/comments in line

TorstenD-SAP (approved):
lgtm

PR #888, opened 2024-02-13
Allow new version of golang linter

halamix2:
**Description**

- Allow new version of golang linter, which uses Node 20

**Related issue(s)**

### DIFF
--- a/docs/contributing/assets/allowed_actions.json
+++ b/docs/contributing/assets/allowed_actions.json
@@ -4,7 +4,10 @@
       "name": "golangci/golangci-lint-action",
       "versions": [
         "639cd343e1d3b897ff35927a75193d57cfcba299",
-        "v3"
+        "v3",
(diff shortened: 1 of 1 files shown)


docs/contributing/assets/allowed_actions.json
```diff
@@ -4,7 +4,10 @@
       "name": "golangci/golangci-lint-action",
       "versions": [
         "639cd343e1d3b897ff35927a75193d57cfcba299",
-        "v3"
+        "v3",
+        "3cfe3a4abbb849e10058ce4af15d205b6da42804",
+        "v4",
```
TorstenD-SAP, line 9: Remove this entry because it allows all minor versions of v4. If not really necessary I am also in favor of removing the next line and only use the commit hash of the release.

halamix2: I've removed `v4`. I'd like to keep `v4.0.0` to allow dependabot to bump it in the future

PR #968, opened 2025-02-12
Update allowed_actions.json with helm/chart-releaser-action

tmilos77:
**Description**

- Add https://github.com/helm/chart-releaser-action/releases/tag/v1.7.0 into allowed_actions.json

**Related issue(s)**


TorstenD-SAP (changes requested):
I scanned the action as well as the chart-releaser tool with Trivy. No vulnerabilities were found. This is why you I suggest to change the `security_review_performed` value to true.

TorstenD-SAP (approved):

PR #1028, opened 2026-05-21
docs: add shell commands guidelines

grego952:
Add section describing how to format shell commands in documentation:
- Distinguish between runnable commands and command-plus-output transcripts
- Use environment variables for copy-paste compatibility
- Follow proper prefix conventions (no prefix for runnable, $ for input lines)

**Description**

- add shell commands guidelines to formatting.md

**Related issue(s)**

### DIFF
--- a/docs/guidelines/content-guidelines/03-formatting.md
+++ b/docs/guidelines/content-guidelines/03-formatting.md
@@ -43,12 +43,15 @@ Custom resources (if you want to refer to the code specifically) | Define the **
 
 ### Shell Commands
 
-When you document shell commands, follow these guidelines:
 
-- Use environment variables instead of hardcoded values so that readers can copy-paste and execute the command without editing. Omit the `export` statements unless the variable value itself is noteworthy — the target audience knows how to define environment variables.
-- Distinguish between the following cases:
-  - **Runnable command** (a snippet for users to copy and execute): Show the command alone, with no prompt prefix and no other `PS1` elements (no `$`, `#`, path, hostname, and so on). The reader must be able to copy the block verbatim into a shell.
-  - **Command-plus-output transcript** (an example showing what the command prints): Prefix only the command lines with `$ ` to mark them as input; leave output lines unprefixed. Omit all other `PS1` elements, such as the path or hostname.
+Use environment variables instead of hardcoded values so that readers can copy-paste and execute the command without editing. Omit the `export` statements unless the variable value itself is noteworthy — the target audience knows how to define environment variables.
+Distinguish between the following cases:
+- **Runnable command** (a snippet for users to copy and execute): Show the command alone, with no prompt prefix and no other `PS1` elements (no `$`, `#`, path, hostname, and so on). The reader must be able to copy the block verbatim into a shell.
+  
+  
+- **Command-plus-output transcript** (an example showing what the command prints): Prefix only the command lines with `$ ` to mark them as input; leave output lines unprefixed. Omit all other `PS1` elements, such as the path or hostname.
+
+  
 
 **Runnable command — no prefix:**
 
--- a/docs/guidelines/content-guidelines/03-formatting.md
+++ b/docs/guidelines/content-guidelines/03-formatting.md
@@ -75,6 +75,30 @@ Suppose that it is only the part containing the certificate information that is 
    }
    ```
 
+## Shell Commands
+
+When shell commands appear in the documentation, follow these guidelines:
+
+- Use environment variables instead of hardcoded values so that the command can be copy-pasted and executed without editing. You can omit the `export` statements unless there is something noteworthy about the variable — the target audience knows how to define environment variables.
+- Distinguish between two cases:
+    - **Runnable command** (a snippet meant to be copy-pasted and executed): show the command alone, with no prompt prefix and no other `PS1` elements (no `$`, `#`, path, hostname, etc.). The reader must be able to copy the block verbatim into a shell.
+    - **Command-plus-output transcript** (an example showing what the command prints): prefix only the command lines with `$ ` to mark them as input; leave output lines unprefixed. Still omit every other `PS1` element (path, hostname, etc.).
+
+**Runnable command — no prefix:**
+
+```
+kubectl -n $NAMESPACE get pods -l app=$APP_LABEL
+```
+
+**Command-plus-output transcript — `$ ` only on the input line:**
+
+```
+$ kubectl -n kyma-system get pods
+NAME                          READY   STATUS    RESTARTS   AGE
+istiod-7c8b9f4d4f-abcde       1/1     Running   0          12d
+kyma-operator-5d9f8b7c-xyz12  1/1     Running   0          5d
+```
+
 ## Panels
 
 >**NOTE** The following panels are not rendered correctly when indented. Use them only in the docs that are displayed on the website.
--- a/docs/guidelines/content-guidelines/03-formatting.md
+++ b/docs/guidelines/content-guidelines/03-formatting.md
@@ -39,6 +39,32 @@ Custom resources (if you want to refer to the code specifically) | Define the **
 >**NOTE:** When you mention specific configuration files in your documents, consider linking to them instead of just mentioning their names. When you link to a file, use its name without the format extension. See the following example:
 > `To adjust the number of Pods in your Deployment, edit the [deployment](./deployment.yaml) file.`
 
+## Code Blocks
+
+### Shell Commands
+
+When you document shell commands, follow these guidelines:
+
+- Use environment variables instead of hardcoded values so that readers can copy-paste and execute the command without editing. Omit the `export` statements unless the variable value itself is noteworthy — the target audience knows how to define environment variables.
+- Distinguish between the following cases:
+  - **Runnable command** (a snippet for users to copy and execute): Show the command alone, with no prompt prefix and no other `PS1` elements (no `$`, `#`, path, hostname, and so on). The reader must be able to copy the block verbatim into a shell.
+  - **Command-plus-output transcript** (an example showing what the command prints): Prefix only the command lines with `$ ` to mark them as input; leave output lines unprefixed. Omit all other `PS1` elements, such as the path or hostname.
+
+**Runnable command — no prefix:**
+
+```
+kubectl -n $NAMESPACE get pods -l app=$APP_LABEL
+```
+
+**Command-plus-output transcript — `$ ` only on the input line:**
+
+```
+$ kubectl -n kyma-system get pods
+NAME                          READY   STATUS    RESTARTS   AGE
+istiod-7c8b9f4d4f-abcde       1/1     Running   0          12d
+kyma-operator-5d9f8b7c-xyz12  1/1     Running   0          5d
+```
+
 ### Omission in Code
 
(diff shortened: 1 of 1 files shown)


docs/guidelines/content-guidelines/03-formatting.md
```diff
@@ -75,6 +75,30 @@ Suppose that it is only the part containing the certificate information that is
    }
    ```
 
+## Shell Commands
+
+When shell commands appear in the documentation, follow these guidelines:
```
NHingerl:
```suggestion
When you document shell commands, follow these guidelines:
```

docs/guidelines/content-guidelines/03-formatting.md
```diff
@@ -75,6 +75,30 @@ Suppose that it is only the part containing the certificate information that is
...
+
+- Use environment variables instead of hardcoded values so that the command can be copy-pasted and executed without editing. You can omit the `export` statements unless there is something noteworthy about the variable — the target audience knows how to define environment variables.
```
NHingerl:
```suggestion
- Use environment variables instead of hardcoded values so that readers can copy-paste and execute the command without editing. Omit the `export` statements unless the variable value itself is noteworthy — the target audience knows how to define environment variables.
Reason
```

docs/guidelines/content-guidelines/03-formatting.md
```diff
@@ -75,6 +75,30 @@ Suppose that it is only the part containing the certificate information that is
...
+- Distinguish between two cases:
+    - **Runnable command** (a snippet meant to be copy-pasted and executed): show the command alone, with no prompt prefix and no other `PS1` elements (no `$`, `#`, path, hostname, etc.). The reader must be able to copy the block verbatim into a shell.
```
NHingerl:
```suggestion
  - **Runnable command** (a snippet for users to copy and execute): Show the command alone, with no prompt prefix and no other `PS1` elements (no `$`, `#`, path, hostname, and so on). The reader must be able to copy the block verbatim into a shell.
```

docs/guidelines/content-guidelines/03-formatting.md
```diff
@@ -75,6 +75,30 @@ Suppose that it is only the part containing the certificate information that is
...
+    - **Command-plus-output transcript** (an example showing what the command prints): prefix only the command lines with `$ ` to mark them as input; leave output lines unprefixed. Still omit every other `PS1` element (path, hostname, etc.).
```
NHingerl:
```suggestion
    - **Command-plus-output transcript** (an example showing what the command prints): Prefix only the command lines with `$ ` to mark them as input; leave output lines unprefixed. Omit all other `PS1` elements, such as the path or hostname.
```

docs/guidelines/content-guidelines/03-formatting.md
```diff
@@ -75,6 +75,30 @@ Suppose that it is only the part containing the certificate information that is
    }
    ```
 
+## Shell Commands
+
+When shell commands appear in the documentation, follow these guidelines:
+
+- Use environment variables instead of hardcoded values so that the command can be copy-pasted and executed without editing. You can omit the `export` statements unless there is something noteworthy about the variable — the target audience knows how to define environment variables.
+- Distinguish between two cases:
```
NHingerl:
```suggestion
- Distinguish between the following cases:
```

NHingerl (commented):
the section above, "Omissions in Code", seems very related. Both are about code blocks (as opposed to single words in code font)
 
1. Does the order make sense - first guidelines about a specific element in code snippets, then guidelines about code snippets in general?
2. Maybe we could integrate them? Is "Omissions in Code" well placed under "Code Font" now that we have "Shell Commands"?

grego952:
> the section above, "Omissions in Code", seems very related. Both are about code blocks (as opposed to single words in code font)
> 
> 1. Does the order make sense - first guidelines about a specific element in code snippets, then guidelines about code snippets in general?
> 2. Maybe we could integrate them? Is "Omissions in Code" well placed under "Code Font" now that we have "Shell Commands"?

For me, Omissions in Code and Shell Commands, even though similar, are two different things. Shell commands are only your input, and "Omissions in Code" is a much broader term. I'd move Omissions in Code under Shell Commands as a separate section. WDYT?

docs/guidelines/content-guidelines/03-formatting.md
```diff
@@ -39,6 +39,32 @@ Custom resources (if you want to refer to the code specifically) | Define the **
 >**NOTE:** When you mention specific configuration files in your documents, consider linking to them instead of just mentioning their names. When you link to a file, use its name without the format extension. See the following example:
 > `To adjust the number of Pods in your Deployment, edit the [deployment](./deployment.yaml) file.`
 
+## Code Blocks
+
+### Shell Commands
+
+When you document shell commands, follow these guidelines:
```
NHingerl: adds no value actually

```suggestion
```

docs/guidelines/content-guidelines/03-formatting.md
```diff
@@ -39,6 +39,32 @@ Custom resources (if you want to refer to the code specifically) | Define the **
...
+
+- Use environment variables instead of hardcoded values so that readers can copy-paste and execute the command without editing. Omit the `export` statements unless the variable value itself is noteworthy — the target audience knows how to define environment variables.
```
NHingerl:
```suggestion
Use environment variables instead of hardcoded values so that readers can copy-paste and execute the command without editing. Omit the `export` statements unless the variable value itself is noteworthy — the target audience knows how to define environment variables.
```

docs/guidelines/content-guidelines/03-formatting.md
```diff
@@ -39,6 +39,32 @@ Custom resources (if you want to refer to the code specifically) | Define the **
...
+- Distinguish between the following cases:
```
NHingerl:
```suggestion
Distinguish between the following cases:
```

docs/guidelines/content-guidelines/03-formatting.md
```diff
@@ -39,6 +39,32 @@ Custom resources (if you want to refer to the code specifically) | Define the **
...
+  - **Runnable command** (a snippet for users to copy and execute): Show the command alone, with no prompt prefix and no other `PS1` elements (no `$`, `#`, path, hostname, and so on). The reader must be able to copy the block verbatim into a shell.
```
NHingerl: What if we add the example here?

```suggestion
- **Runnable command** (a snippet for users to copy and execute): Show the command alone, with no prompt prefix and no other `PS1` elements (no `$`, `#`, path, hostname, and so on). The reader must be able to copy the block verbatim into a shell.
  
  ```
  kubectl -n $NAMESPACE get pods -l app=$APP_LABEL
  ```
```

docs/guidelines/content-guidelines/03-formatting.md
```diff
@@ -39,6 +39,32 @@ Custom resources (if you want to refer to the code specifically) | Define the **
...
+  - **Command-plus-output transcript** (an example showing what the command prints): Prefix only the command lines with `$ ` to mark them as input; leave output lines unprefixed. Omit all other `PS1` elements, such as the path or hostname.
+
+**Runnable command — no prefix:**
```
NHingerl: let's add the examples to the bullet points above

docs/guidelines/content-guidelines/03-formatting.md
```diff
@@ -39,6 +39,32 @@ Custom resources (if you want to refer to the code specifically) | Define the **
...
+## Code Blocks
+
+### Shell Commands
+
+When you document shell commands, follow these guidelines:
+
+- Use environment variables instead of hardcoded values so that readers can copy-paste and execute the command without editing. Omit the `export` statements unless the variable value itself is noteworthy — the target audience knows how to define environment variables.
+- Distinguish between the following cases:
+  - **Runnable command** (a snippet for users to copy and execute): Show the command alone, with no prompt prefix and no other `PS1` elements (no `$`, `#`, path, hostname, and so on). The reader must be able to copy the block verbatim into a shell.
+  - **Command-plus-output transcript** (an example showing what the command prints): Prefix only the command lines with `$ ` to mark them as input; leave output lines unprefixed. Omit all other `PS1` elements, such as the path or hostname.
```
NHingerl: Let's add the example right here.

```suggestion
- **Command-plus-output transcript** (an example showing what the command prints): Prefix only the command lines with `$ ` to mark them as input; leave output lines unprefixed. Omit all other `PS1` elements, such as the path or hostname.

  ```
  $ kubectl -n kyma-system get pods
  NAME                          READY   STATUS    RESTARTS   AGE
  istiod-7c8b9f4d4f-abcde       1/1     Running   0          12d
  kyma-operator-5d9f8b7c-xyz12  1/1     Running   0          5d
  ```
```

NHingerl: suggestions format didn't cope well with the backticks...

docs/guidelines/content-guidelines/03-formatting.md
grego952: Good call! Added

NHingerl (approved):

docs/guidelines/content-guidelines/03-formatting.md
```diff
@@ -39,6 +39,29 @@ Custom resources (if you want to refer to the code specifically) | Define the **
...
+
+
+Use environment variables instead of hardcoded values so that readers can copy-paste and execute the command without editing. Omit the `export` statements unless the variable value itself is noteworthy — the target audience knows how to define environment variables.
+Distinguish between the following cases:
+
+- **Runnable command** (a snippet for users to copy and execute): Show the command alone, with no prompt prefix and no other `PS1` elements (no `$`, `#`, path, hostname, and so on). The reader must be able to copy the block verbatim into a shell. For example:
+
+   ```
+   kubectl -n $NAMESPACE get pods -l app=$APP_LABEL
+   ```
```
IwonaLanger, line 54: I'm not sure if I got it right, but doesn't the example with `$` contradict the description of the runnable command?

gczuczy: The `$` has two uses in shell:

1. In the prompt (the prompt's format is held in the `PS1` environment variable) it can be at the very end, denoting that the session is being run as a user (meaning, it's not the superuser), as the superuser has `#` at the end. This implicitly tells the required permission for the command. Traditionally for a user it's something like `username@hostname:path$ `, telling it's a user named `username`, on the machine called `hostname` and on the filesystem at `path`, and `$` at the end says it's a normal user.
2. Inside a command it denotes a variable, in the above example `$NAMESPACE` and `$APP_LABEL` are variables. The significance of this format is that people are able to simply set the variable to their expected value, and they can run the commands directly without having to modify anything it. Some commands can be longer, and editing them in the terminal is simply uncomfortable - it is much easier to use a notation instead of marked placeholders, what also works natively in the shell for substituting values.

Why the distinction, when the prompt is part of the example, you can't simply copy-paste-run it, and with the usual "copy-to-clipboard" buttons on the renditions, this is expected to work.

This makes the difference for the transcript version, where instead of telling a command to do something, that shows an example of something working in some way - the output there is just as important, and having the prompt (the leading `$` sign, part of PS1) tells the reader what was the command production the output, and which part is the output: it makes a distinction.

IwonaLanger (approved):

